Issue #224, opened 2026-04-13
Manifest/index write race can orphan manifests across crashes


Jesssullivan:
Migrated from tinyland-inc/tummycrypt#35.

Priority: P2

`engine.rs` writes the manifest and then the index. A crash between those writes orphans the manifest and makes it invisible to `list_remote_index()`.

Expected fix direction:
- make the pointer/update sequence crash-safe
- prove the ordering guarantees with focused tests
- document the intended durability model for manifest/index writes

Jesssullivan:
Implementation update for #224 as of 2026-04-13:

Draft PR #243 is now open from `codex/index-entry-v2-parser`. That branch includes:
- legacy + v2 index-entry parsing with explicit `committed` and `preparing` states
- recovery-aware read/delete/reconcile helpers
- staged-manifest -> preparing-index -> final-manifest -> committed-index publish flow
- deterministic crash-window coverage for crashes after staged write, preparing write, final manifest write, and committed index write

Validation run on the branch:
- `cargo test -p tcfs-sync publish_crash_ --locked`
- `cargo test -p tcfs-sync --lib --test two_device_sync_test --test push_pull_roundtrip --locked`
- `cargo check -p tcfs-cli --locked`

Jesssullivan:
Closing as fixed by merged PR #243 (`59c58ce`).

Why this satisfies #224:
- rel-path publication is now recovery-aware instead of "manifest first, index later"
- writes publish through staged manifest -> preparing index -> final manifest -> committed index
- readers recover `preparing` entries deterministically by either materializing/committing the pending manifest or continuing to serve the current committed manifest
- focused crash-window tests cover crashes after staged write, preparing write, final manifest write, and committed index write
- the intended durability model is documented in `notes/manifest-index-durability-2026-04-13.md` and reflected in the landed `index_entry` v2 model

Remaining related work stays on separate issues rather than keeping #224 open:
- `#231` staged/orphaned object cleanup and longer-tail garbage collection
- `#238` broader failure-injection and retry/recovery proof beyond the focused publish crash windows

PR #243, opened 2026-04-14
[codex] Add recovery-aware manifest/index publish durability

Jesssullivan:
## What changed
- added a shared `index_entry` model/parser that supports legacy text entries plus versioned JSON v2 entries with explicit `committed` and `preparing` states
- routed sync read, delete, reconcile, and publish paths through shared recovery-aware helpers
- moved rel-path publication ownership into `upload_file_with_device()` and publish through staged-manifest -> preparing-index -> final-manifest -> committed-index
- added deterministic crash-window tests covering the staged, preparing, final, and committed publish phases

## Why
Issue #224 tracks the manifest/index crash window where a manifest can be uploaded but never become discoverable by path if the process dies before the index is updated.

## Impact
Path-based visibility now has an explicit durability model: readers can deterministically roll forward or preserve the current committed manifest instead of exposing ambiguous or missing state during publish recovery.

## Validation
- `nix develop . --command cargo fmt --all`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo test -p tcfs-sync publish_crash_ --locked`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo test -p tcfs-sync --lib --test two_device_sync_test --test push_pull_roundtrip --locked`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo check -p tcfs-cli --locked`

Refs #224.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -886,16 +886,8 @@ async fn cmd_push(
             ));
             println!("  skipped (unchanged since last sync)");
         } else {
-            // Write index entry for FUSE discoverability (same pattern as push_tree_with_device)
-            let index_key = format!("{}/index/{}", remote_prefix.trim_end_matches('/'), &rel);
-            let index_entry = format!(
-                "manifest_hash={}\nsize={}\nchunks={}\n",
-                result.hash, result.bytes, result.chunks
-            );
-            if let Err(e) = op.write(&index_key, index_entry.into_bytes()).await {
-                eprintln!("warning: failed to write index entry: {e}");
-            }
-
+            // Path publication is handled inside upload_file_with_device so the
+            // manifest/index sequence remains crash-aware.
             pb.finish_with_message("done".to_string());
             println!("  hash:    {}", &result.hash[..16.min(result.hash.len())]);
             println!("  chunks:  {}", result.chunks);
--- a/crates/tcfs-sync/src/engine.rs
+++ b/crates/tcfs-sync/src/engine.rs
@@ -17,7 +17,7 @@ use std::path::{Path, PathBuf};
 use tracing::{debug, info, warn};
 
 use crate::conflict::{compare_clocks, SyncOutcome};
-use crate::index_entry::parse_index_entry;
+use crate::index_entry::{parse_index_entry, RemoteIndexEntry};
 use crate::manifest::SyncManifest;
 use crate::state::{make_sync_state_full, StateCache};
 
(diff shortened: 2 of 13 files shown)


Jesssullivan:
Follow-up on the initial CI failure:

The first `Build + Lint + Test` run failed in `e2e_conflict_resolution` after `#243` made uploads conflict-aware by current `rel_path` index entries. The failure was in the test setup, not the durability path itself: two workflow tests were still modeling pre-index behavior and seeded post-resolution state incorrectly.

Pushed fix: `272128a` (`test(sync): fix conflict resolution e2e setup`)

Local validation after the fix:
- `cargo test -p tcfs-sync --test e2e_conflict_resolution --locked`
- `cargo test -p tcfs-sync --lib --test two_device_sync_test --test push_pull_roundtrip --test e2e_conflict_resolution --locked`

GitHub Actions is rerunning against the updated branch tip now.

Jesssullivan:
Pushed `82cd587` to address the remaining `Build + Lint + Test` failure.

Root cause:
- the durability branch made upload conflict detection consult the current rel-path index entry
- three `e2e_delete_scenarios` cases were still modeling delete/recreate flows as if removing local state also removed the causal lineage for later same-path uploads
- that made the recreated/modified uploads look like unrelated untracked files against a live remote path, so they were skipped

Fix:
- preserve and tick the prior vclock lineage before the recreated/modified upload in:
  - `delete_and_recreate_same_path`
  - `delete_vs_modify_cross_device_conflict`
  - `recreate_deleted_file_with_different_size`
- this matches how the rest of the vclock-aware E2E suite already models local edits after observing a prior synced state

Local validation:
- `nix develop . --command cargo fmt --all`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo test -p tcfs-sync --test e2e_delete_scenarios --test e2e_conflict_resolution --test two_device_sync_test --test push_pull_roundtrip --locked`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo test -p tcfs-sync publish_crash_ --locked`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo check -p tcfs-cli --locked`

Jesssullivan:
Pushed `2db2bc2` to `codex/index-entry-v2-parser`.

Root cause:
- tracked local edits were being compared against the current `rel_path` index entry before the local device tick was applied
- that made normal follow-up writes look like equal-clock conflicts once `#243` started consulting the current path index
- several e2e tests were also moving to new local paths or wiping state before re-uploading without carrying over lineage, which is no longer compatible with path-aware conflict checks

What changed:
- `upload_file_with_device()` now pre-ticks tracked local lineage before rel-path conflict comparison and avoids double-ticking later in the publish path
- the affected e2e suites now explicitly seed lineage when a test changes local paths or intentionally drops state before a re-upload
- `two_device_simultaneous_conflict_detection` now asserts the stricter current behavior: the second concurrent writer is rejected with a recorded conflict instead of publishing a second visible winner

Local verification:
- `nix develop . --command cargo fmt --all`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo test -p tcfs-sync --locked`
- `nix develop . --command env CARGO_TARGET_DIR=/tmp/tummycrypt-224-index-v2-target cargo check -p tcfs-cli --locked`

CI should be rerunning on the new head now.

Issue #234, opened 2026-04-13
Add FUSE integration coverage for real mount flows


Jesssullivan:
Migrated from tinyland-inc/tummycrypt#40.

The VFS layer is tested, but the real FUSE boundary still lacks integration coverage for mount, read, write, and verification flows.

Expected output:
- integration tests that exercise the actual FUSE interface
- clear platform assumptions for where these tests run
- guardrails around flaky kernel or runner behavior

Jesssullivan:
Draft PR #258 is up for the first #234 slice.\n\nThis adds a real tcfs-fuse mount roundtrip integration test with explicit runtime guardrails:\n- runs as a real mount/remount roundtrip on Linux\n- skips cleanly when the environment is not Linux or lacks /dev/fuse or fusermount3\n- proves the test surface is now the actual FUSE boundary, not just VFS-level behavior\n\nValidation run:\n- cargo fmt --all --check\n- cargo test -p tcfs-fuse --features fuse --test mount_roundtrip -- --nocapture\n\nOn this macOS machine the test is visible and skips with the expected `linux-only test` reason. The issue stays open for broader mount/read/write coverage and CI policy on where these tests should run.

Jesssullivan:
Merged #258 as the first real FUSE-bound coverage slice.

That PR adds a guarded Linux mount/remount roundtrip test at the actual tcfs-fuse boundary. The issue stays open because the original contract is broader than a single roundtrip: it still calls for deeper read/write/verification coverage and a clearer CI policy for where these tests should run.

Jesssullivan:
Draft PR #260 is up for the next #234 slice.

This expands the real `tcfs-fuse` coverage from a single remount roundtrip into a broader lifecycle flow through the actual FUSE boundary:
- `mkdir` + directory marker verification
- create/write
- `rename` with old/new remote index verification
- remount + read/list verification
- `unlink` + `rmdir` with remote cleanup verification

Validation run:
- `nix develop /Users/jess/git/tummycrypt --command cargo fmt --all --check`
- `nix develop /Users/jess/git/tummycrypt --command cargo test -p tcfs-fuse --features fuse --test mount_roundtrip -- --nocapture`
- `nix develop /Users/jess/git/tummycrypt --command cargo check -p tcfs-fuse --features fuse`

If this lands cleanly, `#234` should be close to done on behavior; the only likely follow-on would be CI policy about where Linux FUSE runners should exercise these guarded tests.

[...]

PR #260, opened 2026-04-15
[codex] Expand FUSE lifecycle coverage across rename and delete

Jesssullivan:
## What changed
- extend the real `tcfs-fuse` mount integration test surface with a second guarded lifecycle test
- exercise `mkdir`, create/write, `rename`, remount verification, `unlink`, and `rmdir` through the actual FUSE boundary
- verify the corresponding remote index and directory-marker objects are created, moved, and removed as the mounted lifecycle changes

## Why
`#258` established the first real mount/remount roundtrip for `tcfs-fuse`, but `#234` still called out broader mount, read, write, and verification flows. The remaining gap was lifecycle behavior across rename and delete, not another VFS-only test.

## Impact
- gives `#234` real FUSE coverage for the core file and directory lifecycle, not just a single-file roundtrip
- keeps the runtime guardrails explicit: these tests only execute on Linux systems with `/dev/fuse` and `fusermount3`, and otherwise skip cleanly
- adds direct remote-state assertions so the test proves the mount behavior and the backing index behavior stay aligned

## Validation
- `nix develop /Users/jess/git/tummycrypt --command cargo fmt --all --check`
- `nix develop /Users/jess/git/tummycrypt --command cargo test -p tcfs-fuse --features fuse --test mount_roundtrip -- --nocapture`
- `nix develop /Users/jess/git/tummycrypt --command cargo check -p tcfs-fuse --features fuse`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4361,7 +4361,7 @@ checksum = "66d4c8b64f049c6721ec8ccec37ddfc3d641c4a7fca57e8f2a89de509c73df39"
 dependencies = [
  "libc",
  "rtoolbox",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
@@ -4470,7 +4470,7 @@ dependencies = [
  "errno",
  "libc",
  "linux-raw-sys 0.4.15",
- "windows-sys 0.61.2",
+ "windows-sys 0.52.0",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1365,7 +1365,7 @@ dependencies = [
  "libc",
  "option-ext",
  "redox_users 0.5.2",
- "windows-sys 0.61.2",
+ "windows-sys 0.59.0",
 ]
 
 [[package]]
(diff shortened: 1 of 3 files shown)


Jesssullivan:
Draft PR #260 is up for the next #234 slice.

This expands the real `tcfs-fuse` coverage from a single remount roundtrip into a broader lifecycle flow through the actual FUSE boundary:
- `mkdir` + directory marker verification
- create/write
- `rename` with old/new remote index verification
- remount + read/list verification
- `unlink` + `rmdir` with remote cleanup verification

Validation run:
- `nix develop /Users/jess/git/tummycrypt --command cargo fmt --all --check`
- `nix develop /Users/jess/git/tummycrypt --command cargo test -p tcfs-fuse --features fuse --test mount_roundtrip -- --nocapture`
- `nix develop /Users/jess/git/tummycrypt --command cargo check -p tcfs-fuse --features fuse`

If this lands cleanly, `#234` should be close to done on behavior; the only likely follow-on would be CI policy about where Linux FUSE runners should exercise these guarded tests.

PR #373, opened 2026-05-18
Scaffold Linux post-install smoke harness (TIN-1422)

Jesssullivan:
## Summary

First concrete step on TIN-1422 (Linux post-install smoke parity with macOS).
Lands the **shape and gate**, not 100% feature parity. Three new files,
all additive:

1. **`scripts/linux-postinstall-smoke.sh`** — Linux analog of
   `scripts/macos-postinstall-smoke.sh`. Installs a .deb/.rpm, starts
   tcfsd via systemd (with foreground fallback), mounts FUSE, gates
   exact-content hydration on `tcfs index inspect` reporting
   `status=visible` (same `require_expected_remote_index` shape as the
   macOS harness), and optionally exercises evict/rehydrate and mutation.
2. **`scripts/test-linux-postinstall-smoke.sh`** — fake-tools regression
   suite mirroring `scripts/test-macos-postinstall-smoke.sh`. Stubs
   tcfs, tcfsd, mountpoint, fusermount3, systemctl, uname so the
   harness's arg parsing, the index-inspect gate (positive + negative),
   and the hydration flow run on any host.
3. **`.github/workflows/linux-postinstall-smoke.yml`** — workflow
   dispatch mirroring `macos-postinstall-smoke.yml`. Same secret shape
   (`TCFS_SMOKE_S3_ENDPOINT`, `TCFS_SMOKE_S3_BUCKET`,
   `TCFS_SMOKE_S3_ACCESS_KEY_ID`, `TCFS_SMOKE_S3_SECRET_ACCESS_KEY`,
   `TCFS_SMOKE_MASTER_KEY_B64`), same input surface
   (`tag`, `package_url`, `package_artifact_run_id`,
   `package_artifact_name`, `runner_label`, `exercise_evict_rehydrate`,
   `exercise_mutation`). Defaults to `ubuntu-24.04`. Derives
   `storage.enforce_tls` from the endpoint scheme the same way the macOS
   workflow does (after 0b1dc0c).

## Design decisions

- **Clean-name VFS**: Linux FUSE exposes files under their real names
  (no `.tc` stub semantics in the mount surface), so the harness uses
  plain `cat` to trigger hydration through the driver's read path —
  no fileproviderctl-style requestDownload nudges needed.
- **systemd-first**: prefers the unit shipped in the .deb
  (`tcfsd.service`), falls back to foreground `tcfsd --config ...` when
  systemd is unavailable or the unit fails to activate (covers
  containerised runners).
- **Harness installs the package**: keeps the workflow lean. The
  workflow only handles secret wiring, fixture prep, config render, and
  artifact download.
- **Seed via `--seed-expected-file`**: the workflow uses the harness's
  built-in seed flow rather than a separate `tcfs push` step, so the CLI
  used to seed is the same one being smoke-tested.

## Explicit TODOs (TIN-1422 follow-up)

- [ ] **Evict/rehydrate semantics on FUSE**: today the scaffold runs
  `tcfs unsync` against the mount path. The right semantics for FUSE
  flows may need to operate on a sync_root path or coordinate with the
  daemon's eviction queue — flagged inline.
- [ ] **Mutation write-back**: the macOS harness relies on
  FileProvider's queued push. On Linux the operator typically runs
  `tcfs push <sync_root>` to publish mutations; this scaffold assumes
  the FUSE write path queues to remote (may need an explicit push step).
- [ ] **Daemon socket polling**: the scaffold leans on `tcfs status` to
  prove the daemon is reachable rather than polling for a specific
  socket path. Add a socket-path poll once the systemd unit's runtime
  directory layout is canonical.
- [ ] **`.rpm` path coverage in CI**: the default workflow downloads the
  amd64 `.deb`. Add an rpm-on-fedora matrix entry once the release flow
  ships a routable RPM artifact.
- [ ] **Conflict-status fixture**: macOS has `--exercise-conflict-status`
  driven through FileProvider enumerator logs; the Linux scaffold does
  not yet have an analog because there's no equivalent enumerator log
  surface. Needs design.

## Test plan

- [x] `bash scripts/test-linux-postinstall-smoke.sh` passes on macOS
  (the fake `uname` lets the harness's Linux guard pass).
- [x] `bash -n scripts/linux-postinstall-smoke.sh` clean.
- [x] `yq` accepts the workflow YAML.
- [ ] Dispatch `Linux Post-Install Smoke` against a real `.deb` artifact
  on a Linux runner once the workflow merges and the
  `tcfs-linux-smoke` environment + secrets are configured.

Refs: TIN-1422.

### DIFF
--- a/.github/workflows/linux-postinstall-smoke.yml
+++ b/.github/workflows/linux-postinstall-smoke.yml
@@ -1,0 +1,370 @@
+name: Linux Post-Install Smoke
+
+# TIN-1422 scaffold: this workflow mirrors macos-postinstall-smoke.yml for
+# the Linux .deb/.rpm path. It downloads a package artifact (either a
+# release asset or a workflow-run artifact), installs it, writes a live
+# config keyed on the smoke secrets, mounts FUSE, and runs
+# scripts/linux-postinstall-smoke.sh to gate hydration via
+# `tcfs index inspect`.
+#
+# Several steps are intentionally minimal compared to the macOS workflow:
+# Linux doesn't need FileProvider plumbing (host app, pluginkit, CloudStorage
+# discovery), and the smoke fixture path lives directly under the FUSE
+# mount. See TIN-1422 for follow-up items.
+
+on:
+  workflow_dispatch:
+    inputs:
+      tag:
+        description: "Release tag to validate (for example v0.13.0)"
+        required: true
+        type: string
+      package_url:
+        description: "Optional direct .deb/.rpm URL to install instead of the release asset for tag"
+        required: false
+        default: ""
+        type: string
+      package_artifact_run_id:
+        description: "Optional Actions run ID containing a package artifact to install instead of the release asset"
+        required: false
+        default: ""
+        type: string
+      package_artifact_name:
+        description: "Artifact name to download when package_artifact_run_id is set"
+        required: false
+        default: "dist-linux-x86_64"
+        type: string
+      runner_label:
+        description: "Runner label (default: ubuntu-24.04). Use a self-hosted Linux label for private endpoints."
+        required: false
+        default: "ubuntu-24.04"
+        type: string
+      exercise_evict_rehydrate:
+        description: "After initial hydration, exercise unsync + rehydrate"
+        required: false
+        default: false
+        type: boolean
+      exercise_mutation:
+        description: "Write a new file through the FUSE mount and verify by remote pull"
+        required: false
+        default: false
+        type: boolean
+
+permissions:
+  contents: read
+  actions: read
+
+jobs:
+  package-postinstall:
+    name: Linux package post-install smoke
+    if: github.repository == 'Jesssullivan/tummycrypt'
+    runs-on: ${{ github.event.inputs.runner_label }}
+    timeout-minutes: 30
+    environment: tcfs-linux-smoke
+    env:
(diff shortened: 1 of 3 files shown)


Jesssullivan:
Reviewed locally. Scaffold is clean and idiomatic — mirrors the macOS harness's argument surface, log-dir layout, and `require_expected_remote_index` gate exactly. Test harness runs green on yoga (`bash scripts/test-linux-postinstall-smoke.sh` → 'linux-postinstall-smoke tests passed').

**One operator note before first dispatch**: the workflow uses `environment: tcfs-linux-smoke` (new) but the smoke secrets (`TCFS_SMOKE_S3_ENDPOINT`, `TCFS_SMOKE_S3_BUCKET`, `TCFS_SMOKE_S3_ACCESS_KEY_ID`, `TCFS_SMOKE_S3_SECRET_ACCESS_KEY`, `TCFS_SMOKE_MASTER_KEY_B64`) live under `tcfs-macos-smoke` today. Two paths:

1. Create env `tcfs-linux-smoke` and copy the 5 secrets (clean isolation between platforms), OR
2. Change the workflow's `environment:` to `tcfs-macos-smoke` (single shared env, less to maintain).

Either is fine; scaffold-only, no CI auto-trigger, so this can be decided at first-dispatch time.

Will auto-merge on CI green per the day's sprint flow.

PR #377, opened 2026-05-18
ci: real-storage live test lane (TIN-1421)

Jesssullivan:
## Summary

First concrete step on **TIN-1421** (Urgent): brings up SeaweedFS + NATS in CI via the project's existing docker-compose stack and runs `cargo test -p tcfs-e2e --test fleet_live` with `TCFS_E2E_LIVE=1`.

This closes the trust-boundary gap from the 2026-05-18 audit: until now, the live-fleet tests only ran on operator-driven runs against neo/honey/yoga, so PRs could land broken sync paths and `ci.yml` would stay green.

## Design choices

- **Compose-managed services** (not GH Actions `services:`) keeps `docker-compose.yml` as the local + CI source of truth.
- **CI override file** removes the NATS CLI flag that newer images reject and gives SeaweedFS writable volume capacity.
- **Subset only**: masters + volume + filer + NATS. Prometheus/Grafana are observability, not the tested path.
- **Localhost forwards**: `TCFS_S3_ENDPOINT=http://127.0.0.1:8333`, `TCFS_NATS_URL=nats://127.0.0.1:4222`.
- **Path-filtered triggers**: crate/test/config/compose/workflow changes only.
- **Explicit bucket pre-create** surfaces auth/config errors before cargo tests.
- **Compose logs on failure** are uploaded for triage.

## Validation

- [x] `~/.cargo/bin/cargo test -p tcfs-e2e --test fleet_live --no-run`
- [x] CI Live Storage run 26066979207: `fleet_live tests against compose-managed SeaweedFS + NATS` passed in 6m34s
- [ ] Remaining standard PR checks green

## Follow-up

- Fault injection / Toxiproxy lane
- Linux postinstall smoke chain can reuse this backend
- Deliberately broken push-path canary after landing to prove the lane fails for the right reason

### DIFF
--- a/.github/workflows/ci-live-storage.yml
+++ b/.github/workflows/ci-live-storage.yml
@@ -45,6 +45,8 @@ env:
   TCFS_S3_ENDPOINT: "http://127.0.0.1:8333"
   TCFS_S3_BUCKET: "tcfs"
   TCFS_NATS_URL: "nats://127.0.0.1:4222"
+  # Keep infra failures cheap in CI; the production default is 300s.
+  TCFS_UPLOAD_CHUNK_TIMEOUT_SECS: "30"
   AWS_ACCESS_KEY_ID: "admin"
   AWS_SECRET_ACCESS_KEY: "admin"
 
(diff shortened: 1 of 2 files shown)


Jesssullivan:
Review pass after the CI peel-back:

- The workflow is doing the right thing for TIN-1421: repo-managed compose stack, narrow CI override for current SeaweedFS/NATS image behavior, explicit S3/NATS health gates, bucket pre-create, and compose logs on failure.
- The test change is also correct. `live_storage_outage_leaves_no_remote_index_and_recovers` asserts remote index publication, so using `upload_file_with_device(..., NEO_DEVICE, Some(rel_path), None)` is the right API; the old manifest-only helper made the recovery assertion false even when upload recovered.
- Validation so far: local `cargo test -p tcfs-e2e --test fleet_live --no-run` passed, and CI Live Storage run 26066979207 passed `fleet_live` against compose-managed SeaweedFS + NATS in 6m34s.

No blocking review findings from this pass. Still waiting on the standard Nix/Nix-CI checks before marking ready/mergeable. After landing, we should run one deliberate broken-push canary to prove the lane fails red for the right class of sync break.

PR #378, opened 2026-05-18
Design recon: tcfs init first-run wizard (TIN-1425)

Jesssullivan:
## Summary

Design recon for the M11 sprint, NOT implementation.

Proposes the UX shape and integration points for a `tcfs init` first-run wizard
that takes a freshly-installed tcfs binary (.pkg / .deb / .rpm / Homebrew / Nix)
from "binaries on disk" to "`tcfs status` reports `storage [ok]`" without any
hand-editing of `config.toml`, `master.key`, or credentials.

See `docs/ops/tcfs-init-wizard-design-2026-05-18.md`. Cites current behavior
with file:line at:

- `scripts/macos-pkg-postinstall.sh:21-49` (LaunchAgent plist install; no
  config seeding)
- `swift/fileprovider/Sources/HostApp/HostApp.swift:194-204` (silent
  `provisionConfig: no config at ...` path)
- `crates/tcfsd/tcfsd.service:9` (Linux unit hard-wired to
  `/etc/tcfs/config.toml`)
- `crates/tcfs-cli/src/main.rs:2775-2890` (current rudimentary `cmd_init`)

## Discussion seed

The doc closes with six **open questions** that need a human call before any
implementation lands. They are the point of this PR:

1. One-command-or-two for new-identity vs. join-existing-fleet (invite) paths
2. TIN-1417 ordering — fall back to placeholder pubkey or block wizard verify?
3. macOS host app: inline SwiftUI sheet vs. shell out to Terminal `tcfs init`
4. Mnemonic display safety model (plain + checkbox vs. retype-N-words gate)
5. Whether to surface passphrase-mode (Argon2id) in the wizard UI at all
6. Whether to keep a "skip storage for now" escape hatch

## Out of scope

- Implementation (this is a design doc only)
- Any code changes
- Per-device pubkey work (TIN-1417 — assumed parallel)

## Test plan

- [ ] Reviewers walk the doc end-to-end and answer the six open questions
- [ ] Confirm cited file:line references match current main
- [ ] Approve flow before any TIN-1425 implementation PR opens

### DIFF
--- a/docs/ops/tcfs-init-wizard-design-2026-05-18.md
+++ b/docs/ops/tcfs-init-wizard-design-2026-05-18.md
@@ -1,0 +1,171 @@
+# `tcfs init` First-Run Wizard — Design Recon (TIN-1425)
+
+**Status:** design recon for M11 sprint planning. Not an implementation plan.
+**Date:** 2026-05-18.
+**Scope:** propose the UX shape, surfaces, and integration points for a first-run
+wizard that takes a freshly-installed tcfs binary to a green `tcfs status` with
+zero file editing. Implementation is out of scope; per-device pubkey work is
+TIN-1417 and is assumed to land first or in parallel.
+
+## 1. Problem statement
+
+After `installer -pkg TCFS-<version>.pkg -target /` on a fresh Mac:
+
+- `scripts/macos-pkg-postinstall.sh:21-49` writes
+  `/Library/LaunchAgents/io.tinyland.tcfsd.plist` and bootstraps the LaunchAgent
+  into the console user's gui domain (`:60-62`).
+- The plist `ProgramArguments` is hard-wired to
+  `exec /usr/local/bin/tcfsd --config "$HOME/.config/tcfs/config.toml" --mode daemon`
+  (`:36`).
+- The postinstall script never creates `~/.config/tcfs/`, never writes a
+  `config.toml`, and never prompts the user for storage credentials. The
+  LaunchAgent therefore loops on missing-config errors. Logs go to
+  `/tmp/tcfsd.stdout.log` and `/tmp/tcfsd.stderr.log` (`:40-43`) where no
+  end-user will ever look.
+- The bundled `TCFSProvider.app` host app (`swift/fileprovider/Sources/HostApp/HostApp.swift:194-204`)
+  attempts to read `~/.config/tcfs/fileprovider/config.json`; when the file is
+  absent it logs `provisionConfig: no config at ...` and silently skips Keychain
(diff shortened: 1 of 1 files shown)


Jesssullivan:
Review note before this leaves draft:

Recommended decisions on the open questions:

1. Split the surfaces: `tcfs init` should create a new identity/fleet; `tcfs enroll <invite>` should join an existing fleet. `tcfs init` can detect an invite paste and hand off, but the join path should be named explicitly.
2. Do not ship invite-based verify on placeholder device keys. Block the join path until TIN-1417 gives the new device a real local keypair and the invite flow stops treating shared secrets as the trust boundary.
3. macOS should use an inline SwiftUI first-run sheet backed by a machine-readable Rust command/API, not a Terminal hand-off. Terminal is fine as a fallback/debug affordance, not the packaged product path.
4. Require mnemonic confirmation beyond a checkbox: retype a small random subset of words. This is annoying once and prevents a much worse recovery failure.
5. Keep passphrase mode as an advanced CLI option for now unless there is a strong product reason to support both recovery models in the first-run UI.
6. Avoid a casual "skip storage" path. If we need it, make it an explicit disabled state with `tcfs init --resume` and make the daemon/FileProvider avoid registering an apparently usable empty filesystem.

Main blocker: this wizard should not make self-enrollment look production-safe before the TIN-1417/TIN-1424 security work lands.

Jesssullivan:
Review note from the sprint pass:

This is a useful design seed and should stay draft until the enrollment/security dependency is settled. My recommended decisions before implementation:

- Split the UX into `tcfs init` for new fleets and `tcfs enroll <invite>` for joining. That keeps self-enrollment from being hidden inside a broad init command.
- Do not ship the wizard against the placeholder `age1-device-<hash>` identity path. Block the join/verify path on TIN-1417 or clearly mark local-only new-fleet init as the only supported mode.
- Prefer a native macOS sheet, but make it drive a Rust-owned machine contract so Swift does not fork the wizard state machine.
- Treat “skip storage for now” as an explicit disabled state with no FileProvider domain registration. A half-configured Finder surface will create false failures.
- For mnemonic confirmation, use a short random-word retype gate. It is cheap and fits the trust boundary better than a checkbox.

Jesssullivan:
Added a decision log in commit `6928d6f`. The first implementation slice should split `tcfs init` from `tcfs enroll <invite>`, block invite/join verify until TIN-1417/TIN-1424 remove placeholder keys/raw-secret invites, use an inline macOS sheet backed by Rust-owned state, require mnemonic confirmation, hide passphrase mode from the default wizard, and keep skip-storage out of the default GUI path.

PR #379, opened 2026-05-18
Design recon: per-device crypto identity (TIN-1417)

Jesssullivan:
## What this is

Design recon only. This PR adds a single document at
`docs/ops/per-device-crypto-identity-design-2026-05-18.md` proposing the
architecture for TIN-1417 (real per-device cryptographic identity + wrapped
per-device subkeys). No code changes. M11 sprint-planning input.

**This is a DRAFT and is not ready for review-to-land.** It is ready for
review-to-align: please respond inline on the open questions below.

## Why now

The 2026-05-18 audit confirmed `cmd_device_enroll` writes a placeholder
`age1-device-{hash[..8]}` string instead of generating a real keypair
(`crates/tcfs-cli/src/main.rs` around line 2950), and `DeviceRegistry::revoke`
only flips a JSON bool while the revoked device retains the shared master key
forever (`crates/tcfs-secrets/src/device.rs:84`). This blocks TIN-1416,
TIN-1418, and TIN-1424. We need alignment on the substrate before any of those
can move.

## Document at a glance

Problem statement, proposed primitives (age X25519 + Ed25519, keychain-backed
secrets), per-device FileKey wrapping model with explicit recipient-set rules,
honest revocation semantics (revocation prevents NEW reads, cannot un-decrypt
old reads), four-phase migration path, full touch-points list, and tests
needed.

## Open questions seeded for discussion

These are the discussion seed. Please reply on the PR (or in line in the doc)
with calls:

1. age vs HPKE — lock to age, or evaluate HPKE first?
2. Keychain coverage — new `tcfs-keychain` crate, fold into `tcfs-secrets`, or
   adopt `keyring-rs`?
3. Revocation propagation — S3 only with documented window, or also NATS?
4. Forward-secrecy default — auto-rotate on revoke, or manual with reminder?
5. Multi-tenant boundary — recipient set scoped per-tenant, per-prefix, or
   per-file?
6. Old plaintext-secret invites — explicit CHANGELOG note or silent fix?
7. Backward-compat window — one minor release, two, or forever for v2
   manifests on pull?

## Not in this PR

No code. No proto changes. No schema migrations. No CHANGELOG entry. Those
land against TIN-1417 after the questions above are answered.

### DIFF
--- a/docs/ops/per-device-crypto-identity-design-2026-05-18.md
+++ b/docs/ops/per-device-crypto-identity-design-2026-05-18.md
@@ -1,0 +1,265 @@
+# Per-Device Cryptographic Identity Design (TIN-1417)
+
+Status: design recon for M11 sprint planning. Not implementation. This document
+proposes the architecture; the user reviews and aligns before any code lands.
+
+Date: 2026-05-18.
+
+## Problem Statement
+
+Today every "enrolled" device on a tcfs fleet holds the same shared master key
+and there is no per-device cryptographic identity in the local enrollment path.
+Concretely:
+
+- `crates/tcfs-cli/src/main.rs::cmd_device_enroll` (around line 2950) stamps a
+  placeholder public key:
+  `format!("age1-device-{}", &blake3::hash(name.as_bytes()).to_hex()[..8])`.
+  No keypair is generated; the secret half does not exist.
+- `crates/tcfs-secrets/src/device.rs` has an `enroll_remote` helper (line 179)
+  that DOES generate a real `age::x25519::Identity` and return the secret to the
+  caller. The local CLI flow does not call it. So we have one usable code path
+  for real keys but the most-used enrollment command sidesteps it.
+- `DeviceRegistry::revoke` (`crates/tcfs-secrets/src/device.rs:84`) sets
+  `revoked: true` in a JSON blob and saves to disk. The revoked device already
+  holds the shared master key, which derives all FileKeys at decrypt time via
+  `tcfs_crypto::unwrap_key` (`crates/tcfs-sync/src/engine.rs:2055`). Cutting it
+  off the registry does not cut it off the data.
+- The manifest schema (`crates/tcfs-sync/src/manifest.rs:35`) carries a single
+  `encrypted_file_key: Option<String>` — one wrap, under the shared master.
+  There is no recipient set, so there is nothing to remove a device from.
+- `tcfs-auth/src/enrollment.rs` ships `EnrollmentInvite` payloads carrying
+  plaintext S3 access keys, secret keys, and the encryption passphrase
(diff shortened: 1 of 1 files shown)


Jesssullivan:
Review note before this leaves draft:

The design is directionally right, especially the honest revocation section. My recommended decisions:

1. Use age X25519 for M11. It is already in tree and gives us one device-recipient format. Do not block Phase 0 on an HPKE evaluation.
2. Add a small keychain abstraction under `tcfs-secrets` first; only split into `tcfs-keychain` if the platform code grows. The headless Linux fallback must be explicit and documented, not silently weaker.
3. Revocation source of truth should be signed registry state in S3, with a NATS advisory for fast propagation. NATS alone is not durable enough.
4. `tcfs device revoke` should not automatically rewrite multi-GB trees. Default to registry revoke + loud warning, with `--rotate-keys <prefix>` / follow-up command for forward secrecy.
5. Recipient sets should start per-prefix, not per-file. Per-file policy can come later if shared folders need it.
6. Treat existing plaintext-secret invites as a security posture note in the release/change log, not as a silent cleanup. It affects operator trust boundaries.
7. Pull support for v2 single-wrap manifests should be effectively indefinite; write support can be phased out after the migration window.

Hard blocker for implementation: do not let TIN-1425 first-run enrollment depend on the current placeholder `age1-device-*` path or invites that embed raw storage/encryption secrets. The signature also needs to cover the full sensitive invite payload, not just metadata.

Jesssullivan:
Review note from the sprint pass:

This design is directionally right and importantly honest about revocation. Recommended decisions before implementation:

- Standardize on age X25519 for recipient encryption for M11; defer HPKE unless a concrete interoperability need appears.
- Keep raw device secret storage inside `tcfs-secrets` first, with a trait boundary for macOS Keychain / Linux secret-service or file fallback / Windows DPAPI. A new crate can happen after the API proves itself.
- Publish revocations through both S3 registry and NATS. S3 is the durable source of truth; NATS is the low-latency invalidation signal.
- Make key rotation manual by default with a loud `--rotate-keys` option and byte estimate. Automatic multi-GB rewrites are too surprising for the current product state.
- Keep v2 single-wrap manifest read compatibility indefinitely for historical data, but stop writing it after an explicit fleet roll-call/cutover flag.
- Treat existing plaintext-secret invites as an explicit security posture note in release/docs, not a silent fix.

Jesssullivan:
Added a decision log in commit `6ece188` so this design is no longer just a bag of open questions. Key calls: age/X25519 for M11, keychain trait in `tcfs-secrets`, S3 registry as canonical revocation authority with NATS advisory only, manual key rotation on revoke, per-prefix recipient sets first, explicit security note for plaintext-secret invites, and indefinite v2 reads after cutover. Also made the invite signature/full-payload coverage and no-raw-secret self-enrollment boundaries hard gates.

PR #393, opened 2026-05-19
ci: add homebrew install smoke lane (TIN-131)

Jesssullivan:
## Summary
- add a dispatch-only Homebrew fresh-install smoke workflow for the generated homebrew-tap formula
- install from the configured tap branch and run scripts/install-smoke.sh against installed tcfs/tcfsd
- capture Homebrew diagnostics as an artifact for the TIN-131/#280 distribution matrix

## Validation
- ruby -ryaml load/shape check for the new workflow
- git diff --check

Refs: TIN-131, #280

### DIFF
--- a/.github/workflows/homebrew-install-smoke.yml
+++ b/.github/workflows/homebrew-install-smoke.yml
@@ -1,0 +1,78 @@
+name: Homebrew Install Smoke
+
+# TIN-131 / #280: prove the generated homebrew-tap formula installs runnable
+# tcfs binaries from the published release artifacts.
+
+on:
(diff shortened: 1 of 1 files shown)


Jesssullivan:
Ready-for-review note:

- YAML parse and git diff --check are clean.
- I attempted to dispatch the new workflow from the branch, but GitHub Actions returned `workflow not found on the default branch`.
- That is expected for a brand-new workflow file; first live Homebrew smoke can run after this workflow lands on `main`.

Until then this PR proves workflow shape, not Homebrew first-use behavior.

PR #396, opened 2026-05-19
cli: add scoped storage posture canary (TIN-1546)

Jesssullivan:
## Summary
- Add `tcfs storage canary` to run a scoped object-store probe under the configured remote prefix.
- The canary writes one small object, reads exact bytes back, deletes it, verifies it is gone, and reports timings plus TLS/enforce_tls posture.
- Support `--json` so evidence packets can archive machine-readable endpoint/bucket/prefix/key/timing data without secrets.

## Claim boundary
This adds a canary tool and memory-backed regression coverage. It does not by itself prove production S3 posture. TIN-1546 still needs a live production-like HTTPS/CA packet with scoped credentials, transient classification, and large restore/object-count evidence.

## Validation
- `~/.cargo/bin/cargo fmt --all -- --check`
- `~/.cargo/bin/cargo test -p tcfs-cli storage_canary -- --nocapture`
- `git diff --check`

TIN-1546

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -116,6 +116,12 @@ enum Commands {
         action: IndexAction,
     },
 
+    /// Storage posture checks
+    Storage {
+        #[command(subcommand)]
+        action: StorageAction,
+    },
+
     // ── Phase 3: mount + stub management ──────────────────────────────────────
     /// Mount a remote as a local directory
     Mount {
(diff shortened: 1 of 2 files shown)


Jesssullivan:
Added a follow-up docs commit (`8c8ccd7`) tying `tcfs storage canary --json` into the alpha/beta evidence packet checklist. Claim boundary remains unchanged: the command provides scoped read/write/delete/delete-verify evidence, but TIN-1546 still needs live HTTPS/CA, scoped credentials, transient classification, and large restore/object-count proof.

PR #428, opened 2026-05-21
fileprovider: preserve manifests during rename safety proof (TIN-1547)

Jesssullivan:
## Summary
- adds a macOS postinstall harness flag to prove FileProvider rename safety through the CloudStorage/Finder path
- verifies destination remote index/content after rename and verifies the old remote path is no longer visible
- fixes the rename root cause: deleting the old path now removes the index entry first and preserves content-addressed manifests still referenced by another visible path
- routes FileProvider delete paths through the shared sync-engine primitive so direct, gRPC, and UniFFI callers use the same manifest-preservation behavior
- wires the rename gate into `macos-postinstall-smoke.yml` as an optional post-M10 hardening input

## Evidence So Far
- Previous signed-package smoke against rc3 artifact reached base hydrate, evict/rehydrate, and mutation, then failed the new rename gate because `rename-...-renamed.txt` had a visible index entry pointing at a missing manifest.
- Root cause: FileProvider rename copied/uploaded the destination, then deleted the old path; the delete removed the shared content-addressed manifest object that the destination index still referenced.

## Verification
- `bash -n scripts/macos-postinstall-smoke.sh`
- `bash scripts/test-macos-postinstall-smoke.sh`
- `bash scripts/test-release-workflow-fileprovider.sh`
- `git diff --check`
- `~/.cargo/bin/cargo fmt --all -- --check`
- `~/.cargo/bin/cargo test -p tcfs-sync delete_remote_file`
- `~/.cargo/bin/cargo test -p tcfs-file-provider delete_one_duplicate_content_path_preserves_other_path`
- `~/.cargo/bin/cargo test -p tcfs-file-provider --no-default-features --features grpc master_key_reads_base64_config`
- `~/.cargo/bin/cargo check -p tcfs-file-provider --no-default-features --features uniffi`

## Tracker
Advances TIN-1547. This should stay draft until the fresh signed/notarized package artifact passes the layered macOS postinstall smoke with `--exercise-rename-safety`.


Jesssullivan:
TIN-1547 update: the signed rc3 package smoke reached the new rename gate and failed in the product path, not setup. Run https://github.com/Jesssullivan/tummycrypt/actions/runs/26205119069 passed install/signing/storage/seed/base hydrate/evict-rehydrate/mutation, then failed after rename because the destination index was committed but its content-addressed manifest had been deleted by the old-path delete.\n\nRoot cause: FileProvider rename copies/uploads the new path and then deletes the old path; delete removed the committed manifest object even when another index entry still referenced the same manifest hash.\n\nFix now pushed in `a572726`: route FileProvider delete through a shared sync-engine delete primitive that preserves manifests referenced by other index entries; added sync and FileProvider direct regressions. Local verification passed:\n- `cargo fmt --all -- --check`\n- `git diff --check`\n- `cargo test -p tcfs-sync delete_remote_file`\n- `cargo test -p tcfs-file-provider delete_one_duplicate_content_path_preserves_other_path`\n- `cargo test -p tcfs-file-provider --no-default-features --features grpc master_key_reads_base64_config`\n- `cargo check -p tcfs-file-provider --no-default-features --features uniffi`\n\nNext proof in flight: rebuild a signed/notarized package from this PR branch, then rerun the same rename-safety smoke against that fresh artifact.

Jesssullivan:
Fresh signed-package proof is green.

- Package proof run: https://github.com/Jesssullivan/tummycrypt/actions/runs/26205517985
- Package artifact: `dist-notarized-pkg-proof` / artifact id `7127387410`
- Postinstall smoke run: https://github.com/Jesssullivan/tummycrypt/actions/runs/26205772566
- Smoke artifact: `macos-postinstall-smoke-v0.12.13-rc3` / artifact id `7127417491`

What the smoke proved from the #428 branch package artifact:
- package install and installed FileProvider signing
- installed-binary smoke
- live config + FileProvider config provisioning
- seeded remote fixture and E2EE negative/positive gate
- daemon storage `[ok]`
- signed HostApp/root probe path
- exact hydrate
- evict/rehydrate
- mutation remote pull
- rename safety: destination path remains readable from remote and old path is no longer visible

This verifies the fix for the previous failure mode where rename copied/uploaded the destination but delete of the old path removed a content-addressed manifest still referenced by the destination index.

PR #494, opened 2026-06-07
DRAFT: TIN-1417 B3 — restore Phase-1 dual-write + per_device_wrap_strict flag (agent-drafted, needs review)

Jesssullivan:
> **DRAFT — agent-drafted, needs review.** Cut from `origin/main`. Does NOT flip any flag; both `crypto.per_device_wrapping` and the new `crypto.per_device_wrap_strict` default **false**. Default-off behavior is byte-identical.

## TRACK B3 — Step 2: restore Phase-1 dual-write + the `per_device_wrap_strict` flag model

The current `origin/main` per-device write path is a **clean-cut Phase-2**: when `device_recipients` is non-empty it emits `wrapped_file_keys` and **drops** `encrypted_file_key` (the master wrap). That leaves no rollback and is unreadable by old binaries / the master-key-only path / the rotate tool. The TIN-1417 design (`docs/ops/per-device-crypto-identity-design-2026-05-18.md:207-216`) specifies **Phase-1 DUAL-WRITE** (both fields) plus a roll-call gate before contracting.

This PR restores that dual-write as the default and gates the contraction behind a new, default-off flag.

### Flag semantics
- `crypto.per_device_wrapping` (**EXPAND**, unchanged default `false`): when true, also wrap the file key to per-device recipients (emit `wrapped_file_keys`).
- `crypto.per_device_wrap_strict` (**CONTRACT**, NEW, default `false`): only meaningful when `per_device_wrapping` is true. When **false (default)** → DUAL-WRITE: keep `encrypted_file_key` (master wrap, the rollback fallback) **and** `wrapped_file_keys`. When **true** → drop the master wrap (today's clean-cut). Strict is only safe **after a fleet roll-call** confirms every active device has a real per-device identity enrolled — and it is unreadable by old binaries / the master-key path.

So: default (`strict=false`) = dual-write; `strict=true` = current clean-cut.

### Changed files
- `crates/tcfs-core/src/config.rs` — new field `crypto.per_device_wrap_strict: bool` (default `false`), mirroring `per_device_wrapping`'s serde/default (`CryptoConfig` has `#[serde(default)]`, so existing config files deserialize unchanged).
- `crates/tcfs-sync/src/engine.rs`:
  - `EncryptionContext` gains `strict_device_wrap: bool` (default `false`) + a minimal `with_strict_device_wrap(bool)` builder. `new()` and `with_device_wrapping(..)` keep their existing signatures (no caller churn).
  - The per-device write branch (was lines 1858-1881; the impl is `upload_file_with_device_with_state`) now DUAL-WRITES `encrypted_file_key` + `wrapped_file_keys` unless `strict_device_wrap` is set, in which case it omits the master wrap. Manifest stays **version 2**.
  - The **read switch (engine.rs:2226-2270) is unchanged** — it already prefers `wrapped_file_keys` and falls back to `encrypted_file_key`, so dual-written manifests are readable by both.
- `crates/tcfsd/src/grpc.rs` and `crates/tcfs-cli/src/main.rs` — both copies of `build_encryption_context` thread the config flag through `.with_strict_device_wrap(config.crypto.per_device_wrap_strict)`. (Did NOT depend on the unmerged #492 FP-local copy.)
- `crates/tcfs-sync/tests/per_device_dual_write_test.rs` — NEW.
- `crates/tcfs-sync/tests/per_device_wrap_test.rs` — pinned to strict mode (its clean-cut / revocation intent), with a header pointer to the new dual-write test.

### Dual-write proof (new tests, all passing)
`per_device_dual_write_test.rs`:
- `dual_write_emits_both_fields_and_is_readable_by_both_readers` — with `per_device_wrapping` ON + strict OFF, the written manifest carries **BOTH** `encrypted_file_key` AND `wrapped_file_keys` (and `version == 2`); the master wrap actually unwraps under the master key; **per-device readers** hydrate exact bytes via the per-device switch; and a **master-key-only reader** (no age identity — the old-binary / master-key-direct path, which never consults `wrapped_file_keys`) hydrates exact bytes via the `encrypted_file_key` fallback.
- `strict_mode_omits_master_wrap` — with strict ON, only `wrapped_file_keys` (no master wrap); recipient still hydrates.
- `default_off_path_is_legacy_master_only` — with no recipients (per_device_wrapping off), only `encrypted_file_key`, no `wrapped_file_keys` (byte-identical legacy).

### FULL LOCAL CI GATE
- `cargo fmt --all -- --check` → **clean**.
- `cargo clippy --workspace --all-targets` → **exit 0, no new warnings**. (Two pre-existing `clippy::needless_borrow` warnings at `grpc.rs:1383` and `grpc.rs:1816` are on `origin/main` and untouched by this PR.)
- `cargo test -p tcfs-sync --features crypto` (the touched tests): `per_device_dual_write_test` **3 passed**, `per_device_wrap_test` **2 passed**, `encrypted_roundtrip_test` **4 passed**.
- `cargo test -p tcfs-core`: **8 passed** (incl. config serde round-trip / full-config parse).

### Operator notes / uncertainty
- **The `per_device_wrap_strict` NAME and model need operator ratification.** "strict" = CONTRACT (drop master wrap). If the fleet prefers a different name (e.g. `per_device_wrap_contract`, `drop_master_wrap`) or a tri-state, that is a trivial rename before merge.
- **Read-path nuance worth a second look:** the read switch is intentionally left unchanged per spec. For a *dual-written* manifest it still takes the per-device branch first (because `wrapped_file_keys` is non-empty) and does **not** fall back to the master wrap if a per-device unwrap fails. The master-readability of a dual-written manifest is therefore realized by readers that go through the `encrypted_file_key` path directly (old binaries, the master-key-only/FileProvider-direct path, the rotate tool) — exactly the rollback audience. The new test simulates that reader faithfully. A modern reader that holds *only* a master key but no age identity will still error on a dual-written manifest under the current switch; if operators want belt-and-suspenders master fallback inside the modern read switch too, that is a follow-up (and a read-switch change, out of scope here).
- **Lab spec (not applied):** `~/git/lab/nix/home-manager/tummycrypt.nix` `[crypto]` block sets only `enabled`/`master_key_file`, so both flags already default false there — no lab change is required to keep defaults off. To later EXPAND, add `per_device_wrapping = true` (after enrolling real per-device identities + `device_registry_path`/`device-<id>.age`); to CONTRACT, add `per_device_wrap_strict = true` only after a roll-call. Do NOT set strict before dual-write has propagated.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -1105,6 +1105,11 @@ fn resolve_sync_status_lookup_path(path: &Path) -> Result<PathBuf> {
 /// recipients. Falls back to legacy shared-master wrapping (logging why) if the
 /// registry can't be loaded, has no real recipients, or this device's age secret
 /// is missing — never producing content this device cannot read back.
+///
+/// By default (Phase-1) the write path DUAL-WRITES: per-device `wrapped_file_keys`
+/// plus the master-wrapped `encrypted_file_key` as a rollback fallback. Setting
+/// `crypto.per_device_wrap_strict` flips it to the CONTRACT (clean-cut) behavior
+/// that drops the master wrap — only safe after a fleet roll-call.
 fn build_encryption_context(
     config: &tcfs_core::config::TcfsConfig,
     device_id: &str,
@@ -1156,6 +1161,7 @@ fn build_encryption_context(
         }
     };
     base.with_device_wrapping(recipients, Some(identity))
+        .with_strict_device_wrap(config.crypto.per_device_wrap_strict)
 }
 
 // ── `tcfs push` ───────────────────────────────────────────────────────────────
--- a/crates/tcfs-core/src/config.rs
+++ b/crates/tcfs-core/src/config.rs
@@ -329,12 +329,24 @@ pub struct CryptoConfig {
     /// Generated once per vault. If unset and passphrase_file is used, a random
     /// salt is generated and must be persisted by the caller.
     pub kdf_salt: Option<String>,
-    /// Wrap file keys per-device (age/X25519) for non-revoked devices instead of
-    /// under the shared master key (TIN-1417). When true, new manifests carry
-    /// `wrapped_file_keys` and omit `encrypted_file_key`, so a revoked device
-    /// cannot decrypt newly written content. Default false (legacy shared-master)
-    /// until a fleet has real per-device identities enrolled.
+    /// Wrap file keys per-device (age/X25519) for non-revoked devices in addition
+    /// to the shared master key (TIN-1417). This is the EXPAND switch: when true,
+    /// new manifests carry `wrapped_file_keys`. By default it remains a Phase-1
+    /// DUAL-WRITE — manifests ALSO keep the master-wrapped `encrypted_file_key`
+    /// as a rollback fallback readable by old binaries and the master-key path —
+    /// so per-device readers and master-key readers can both decrypt. Default
+    /// false (legacy shared-master) until a fleet has real per-device identities
+    /// enrolled. See `per_device_wrap_strict` to CONTRACT (drop the master wrap).
     pub per_device_wrapping: bool,
+    /// Strict per-device wrapping: the CONTRACT switch (TIN-1417 Phase-2). Only
+    /// meaningful when `per_device_wrapping` is true. When true, new manifests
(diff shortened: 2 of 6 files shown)


Jesssullivan:
Superseded by the ratified tri-state crypto.wrap_mode model (#496). The two-bool per_device_wrapping/per_device_wrap_strict design is replaced by the single wrap_mode enum. Closing.

PR #508, opened 2026-06-08
DRAFT: reconcile restore fidelity (mtime preserve + symlink opt-in) for zero-diff roam — TIN-1620 T13-Z (agent-drafted, needs review)

Jesssullivan:
DRAFT — agent-drafted, needs human review before merge. References **TIN-1620** (Phase-2 / G5 repo-roam) precision sub-row **T13-Z** (dev-env zero-diff roam).

## Goal

Enroll a `~/git` repo so it roams neo<->honey with **zero dev-env difference** (`git status` clean; uncommitted+staged+untracked+branch+stash all follow) via the scheduled `tcfs reconcile --execute` extraReconcileRoots unit pointed at a per-repo config with `sync_git_dirs=true` + `git_sync_mode=raw`. This PR fixes the **two restore-fidelity blockers** that otherwise make `git status` report spurious-dirty after a roam.

## BLOCKER A — preserve mtime on restore

A fresh restore stamps "now" onto every file, smudging `.git/index`'s stat cache so `git status` re-hashes everything and reports spurious-dirty.

- **Manifest field** (`crates/tcfs-sync/src/manifest.rs`): added optional `mtime: Option<(i64, u32)>` (unix secs + subsec nanos), `#[serde(default, skip_serializing_if = "Option::is_none")]`.
- **Capture** (`crates/tcfs-sync/src/engine.rs`, upload path): captured from the **same** `std::fs::Metadata` read that already yields `file_mode` (single stat, no TOCTOU drift), via `meta.modified()`.
- **Apply** (`crates/tcfs-sync/src/engine.rs`, restore paths): after the atomic temp-file rename and **before** `make_sync_state_full` re-stats, via `libc::utimensat` under `#[cfg(unix)]` (`UTIME_OMIT` for atime). Applied on **both** the empty-file and chunked-file restore branches. `mtime: None` (old manifests) leaves current behavior unchanged. Used `libc` (already a dep) rather than adding the `filetime` crate, to minimize new surface.

### Manifest back-compat analysis (identity/addressing)

Manifest identity/addressing is computed from **file content only**, never the manifest's serialized bytes:

- The manifest object key is `manifests/{file_hash}`, where `file_hash` = `tcfs_chunks::hash_to_hex(<BLAKE3 of file content>)` (`engine.rs`). The index entry's `manifest_hash` is set to that same `file_hash_hex` (`RemoteIndexEntry::new(file_hash_hex, …)`).
- Dedup is keyed purely on `op.exists("…/manifests/{file_hash}")`; the existing manifest is read only to count chunks, never byte-compared.
- Chunks are content-addressed by their own BLAKE3.

Therefore adding `mtime` **cannot** change manifest identity, dedup, or chunk addressing — the fleet will not re-upload or re-address existing content. This is the same property the existing `mode` field already relies on. Proven by a test that an `mtime: None` manifest serializes **byte-identically** to a pre-field manifest (the key is omitted) and that an old blob with no `mtime` key deserializes to `None`.

## BLOCKER B — opt symlinks in

`crates/tcfs-sync/src/reconcile.rs`:

- `collect_local_set` now sets `preserve_symlinks: true` (keeping `follow_symlinks: false`), and folds `CollectResult::symlinks` into the `rel_path → PathBuf` map so reconcile actually sees them (previously only `result.files` was used, so a `preserve_symlinks` link would have been invisible to reconcile).
- The `execute_plan` Push arm now detects a symlink via `symlink_metadata(...).is_symlink()` and dispatches to `engine::upload_symlink_with_device` (first-class symlink manifest) instead of the chunked-file uploader, mirroring the proven full-tree-upload dispatch. The restore side (`download_file_with_device`) and the fail-closed deny-set guard already handle symlinks.

## Tests (non-vacuous)

- `mtime_round_trips_for_chunked_file` / `mtime_round_trips_for_empty_file` — upload a file stamped at a known mtime, restore into a **fresh** dir, assert restored mtime == source (secs exact, nanos within precision). Covers both restore paths.
- `old_manifest_without_mtime_restores_with_current_behavior` + `mtime_none_is_omitted_and_back_compatible` — a manifest serialized **without** `mtime` deserializes to `None`, restores without panic/error, keeps fresh-write timestamp, and `mtime: None` is omitted from JSON (no addressing change).
- `reconcile_push_then_restore_round_trips_symlink` — a local-only symlink is collected (not dropped/dereferenced), pushed as a **symlink manifest** (asserted, not regular-file content), then restored on a fresh peer as a link with target intact.
- `systemtime_to_unix_parts_*` — pre/post-epoch conversion.

## Scheduled-unit / canary note

The scheduled reconcile unit (or canary) should run `git update-index --refresh -q` **once post-pull** so `.git/index`'s stat cache is rewritten deterministically against the restored mtimes. (The dedicated `docs/ops/repo-roam-test-plan-2026-06-08.md` does not exist on `main`, so this note lives here.)

## Full local CI gate

- `cargo fmt --all -- --check` — **clean**.
- `cargo clippy --workspace --all-targets` — no NEW warnings. The 2 pre-existing `needless_borrow` warnings in `tcfsd/src/grpc.rs:1442,1876` (on `storage_prefix`) are unrelated to this change (my only grpc.rs edit is one `mtime: None,` line). `cargo clippy -p tcfs-sync --all-targets` (and `--features crypto`) is fully clean.
- `cargo test` — green for every touched crate: `tcfs-sync` (208 lib + integration), `tcfs-cli`, `tcfs-file-provider`, `tcfs-vfs`, `tcfsd`.
- cargo-deny: out of scope (handled separately).

## Uncertainties / loud flags

- **Non-Unix mtime restore is a no-op** — the value round-trips through the manifest but `apply_manifest_mtime` only restamps under `#[cfg(unix)]`. The fleet targets are Unix (neo/honey), so this is fine for the goal, but a Windows port would need wiring.
- **`compare_both_exist` symlink edge not expanded** — the new symlink push path is exercised for the NewLocal case (the zero-diff goal). A *tracked* symlink that already exists remotely flows through `compare_both_exist`, which calls `hash_file` (follows the link). I deliberately did not expand scope into that vclock-comparison path; for the enrollment goal the symlink set is published once via NewLocal pushes. Flagging for reviewer attention if bidirectional symlink edits are in scope.
- **`--all-features` clippy/build** could not run in this sandbox (`librocksdb-sys` fails with `empty search path given via -L`, an environment toolchain issue, not a code defect). Default-feature and `--features crypto` gates are clean.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -7380,6 +7380,7 @@ mod tests {
             written_at: 0,
             rel_path: Some(rel_path.to_string()),
             mode: None,
+            mtime: None,
             encrypted_file_key: Some(base64::engine::general_purpose::STANDARD.encode(wrapped)),
             wrapped_file_keys: Vec::new(),
         }
@@ -8552,6 +8553,7 @@ enabled = false
             written_at: 0,
             rel_path: Some(rel_path.to_string()),
             mode: None,
+            mtime: None,
             encrypted_file_key,
             wrapped_file_keys,
         };
@@ -8835,6 +8837,7 @@ enabled = false
             written_at: 0,
             rel_path: Some("plain.txt".into()),
             mode: None,
+            mtime: None,
             encrypted_file_key: None,
             wrapped_file_keys: Vec::new(),
         };
--- a/crates/tcfs-file-provider/src/device_ctx.rs
+++ b/crates/tcfs-file-provider/src/device_ctx.rs
@@ -882,6 +882,7 @@ mod tests {
             written_at: 0,
             rel_path: None,
             mode: None,
+            mtime: None,
             encrypted_file_key: None,
             wrapped_file_keys: Vec::new(),
         };
--- a/crates/tcfs-file-provider/src/direct.rs
+++ b/crates/tcfs-file-provider/src/direct.rs
@@ -722,6 +722,7 @@ pub unsafe extern "C" fn tcfs_provider_upload(
                 written_at,
                 rel_path: Some(remote_str.to_string()),
                 mode: None,
+                mtime: None,
                 encrypted_file_key,
                 wrapped_file_keys: Vec::new(),
             };
@@ -1467,6 +1468,7 @@ mod tests {
                 written_at: 0,
                 rel_path: Some(rel.to_string()),
                 mode: None,
+                mtime: None,
                 encrypted_file_key,
                 wrapped_file_keys,
             };
--- a/crates/tcfs-file-provider/src/uniffi_bridge.rs
+++ b/crates/tcfs-file-provider/src/uniffi_bridge.rs
@@ -593,6 +593,7 @@ impl TcfsProviderHandle {
                 written_at,
                 rel_path: Some(remote_path.to_string()),
                 mode: None,
+                mtime: None,
                 encrypted_file_key,
                 wrapped_file_keys: Vec::new(),
             };
@@ -980,6 +981,7 @@ mod tests {
                 written_at: 0,
                 rel_path: Some(rel.to_string()),
                 mode: None,
+                mtime: None,
                 encrypted_file_key,
                 wrapped_file_keys,
             };
--- a/crates/tcfs-sync/src/engine.rs
+++ b/crates/tcfs-sync/src/engine.rs
@@ -2227,7 +2227,7 @@ pub async fn upload_symlink_with_device(
     })
 }
 
-fn read_symlink_target_text(path: &Path) -> Result<String> {
+pub(crate) fn read_symlink_target_text(path: &Path) -> Result<String> {
     let target = std::fs::read_link(path)
         .with_context(|| format!("reading symlink target: {}", path.display()))?;
     target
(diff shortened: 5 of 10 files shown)


Jesssullivan:
Blocking metadata-model concern from a read-only review:

The mtime/mode restore idea is right, but storing path-specific metadata only inside `SyncManifest` while the committed manifest object is addressed as `manifests/{file_hash}` appears insufficient for zero-diff repo roam.

Code anchors:

- `RemoteIndexEntry` carries only `manifest_hash`, `size`, `chunks`, `kind`, and optional symlink target. It does not carry regular-file mode or mtime.
- `publish_manifest_for_rel_path_with_hook` computes `final_manifest_key = manifests/{entry.manifest_hash}` and skips writing the final manifest if that key already exists.
- For regular files `entry.manifest_hash` is currently the file content hash.

That means two different paths with identical bytes but different mtime or mode both point at the same final manifest object. Whichever path first publishes `manifests/{content_hash}` wins the stored metadata; later duplicate-content paths commit index entries pointing at that same object without updating it. On restore, both paths can receive the same mtime/mode, which defeats the git index/stat-cache goal for common cases like empty files, generated marker files, duplicated fixtures, or same-content scripts with different executable bits.

Before this merges, I think it needs one of:

1. move per-path regular-file metadata into the remote index entry, or
2. make the manifest object key a real manifest identity, e.g. hash of manifest bytes / content+metadata, while keeping `SyncManifest.file_hash` as the plaintext content hash for integrity, or
3. explicitly scope out duplicate-content metadata fidelity and add a failing/xfail test documenting that T13-Z cannot claim full zero-diff under that condition.

A required regression test should cover at least two same-content files with different mtimes and/or modes, then restore both through the path-indexed flow and assert each path gets its own metadata. The current mtime tests only cover a single file per content hash, so they do not exercise this collision.

Jesssullivan:
TIN-1910 fix is staged as stacked PR #510: https://github.com/Jesssullivan/tummycrypt/pull/510

It keeps `UploadResult.hash` as the plaintext/content hash, but path-indexed uploads now publish/index a metadata-sensitive manifest object id. That removes the duplicate-content collision for mode/mtime/rel_path/vclock while preserving chunk dedup.

Validation on #510 branch:
- `~/.cargo/bin/cargo fmt --check`
- `git diff --check`
- focused duplicate-content metadata regression
- committed-index contract test
- `two_device_sync_test`
- `push_pull_roundtrip`
- full `~/.cargo/bin/cargo test -p tcfs-sync -q`

Jesssullivan:
Correction/update: #508 has already merged to main, so #510 has been retargeted from the stack base to main. It is now the direct TIN-1910 regression fix for the path-scoped metadata collision: https://github.com/Jesssullivan/tummycrypt/pull/510

PR #513, opened 2026-06-09
DRAFT: .git-aware fast-forward conflict resolution (closes bidirectional repo-roam R3; G5-git-5 FF half) — agent-drafted, needs review

Jesssullivan:
> Stacked on #512 (cleanup of the tcfs-sync conflict/reconcile chokepoints). Review/merge #512 first; this branch was cut from `origin/cleanup/tcfs-sync-conflict-dead-dry`, so the diff against `main` includes #512's commit until that lands.

Agent-drafted, needs review.

## What

Implements `.git`-aware **fast-forward** conflict resolution for raw git-dir sync, closing the bidirectional repo-roam **R3** case (the common "work on honey, pick up on neo" handoff). This is the **FF half** of G5-git-5. Divergent keep-both (T10/T11) is intentionally left as a separate follow-up — for divergent `.git` states this PR **leaves the `Conflict` as-is** (fail-closed).

## Why bidirectional `.git` roam failed

`compare_clocks` saw each device independently tick its vclock on `.git/refs/heads/<branch>` + `.git/index` + `.git/logs`, so concurrent edits classified as `Conflict` with no git-ancestry check to demote a fast-forward; and the `Conflict` arm of `execute_plan` only *records*, never resolves.

## Design

1. **Reclassifier at the conflict boundary** (`reclassify_git_ff_conflicts` in `reconcile.rs`), run as a post-classification pass — only when `git_ff_resolution` is set **and** `git_sync_mode == "raw"`. Generic `compare_clocks` for normal files is untouched.
2. For a conflicting `.git` path: find the enclosing repo root + the branch ref, read the local commit SHA (live repo) and the remote commit SHA (remote ref blob), and probe ancestry locally via `git merge-base --is-ancestor` (`classify_fast_forward` in `git_safety.rs`):
   - remote ancestor of local → **Push** (LocalNewer)
   - local ancestor of remote → **Pull** (RemoteNewer)
   - neither / equal-but-different / a needed object not yet local → **stays Conflict** (fail-closed / defer)
   - the repo's `.git` conflicts are resolved **atomically** toward one winner; a repo is never split half push / half pull.
3. **Ordering**: in both `execute_plan` (sequential) and the concurrent new-remote pull fast path, `.git/objects/**` + packs apply **before** refs (`.git/refs/**`, `packed-refs`, `HEAD`) so a ref never advances to an object not yet present locally.
4. **TOCTOU**: `acquire_git_lock` (previously **zero callers**) is now wired into `execute_plan` via an RAII `GitLockGuard`, holding a per-repo `.git/tcfs.lock` across the apply window.
5. **Config**: `git_sync_mode` + `git_ff_resolution` threaded through `ReconcileConfig`; an encryption context threaded through `reconcile()` (needed to read remote ref blobs). Wired in the daemon + CLI reconcile paths (FF on when `sync_git_dirs` + `raw`).

## Test

`crates/tcfs-sync/tests/git_ff_resolution.rs`:
- **FF converges**: two devices share a repo at C0; B commits C1 (C0 ancestor of C1); B's reconcile **pushes** (LocalNewer, not Conflict), A's reconcile **pulls** to C1 (RemoteNewer); both end at C1, `git fsck --full` clean both sides.
- **Divergent stays Conflict**: B@C1, A@C2 (neither an ancestor), genuinely concurrent vclocks → the `.git/refs/heads/main` conflict **remains a Conflict** and is never reclassified.

## Gate (local)

- `cargo fmt --all -- --check`: clean
- `cargo clippy --workspace --all-targets`: only the 2 pre-existing `needless_borrow` warnings in `tcfsd/src/grpc.rs`; no new warnings
- `cargo test -p tcfs-sync` (incl. new FF test) + `cargo test -p tcfsd`: all pass


Jesssullivan:
## Deep review (agent-assisted, adjudicated) — verdict: changes-needed

Core ancestry logic verified correct and fail-closed; blast radius verified flag-gated. But two must-fix findings + one barrier gap block merge:

**HIGH-1 — `read_remote_ref_sha` fabricates `.git` at the sync root** (`reconcile.rs:1171`): uses `local_root.join(".git/tcfs-ff-tmp")` where the in-scope `repo_root` was intended. Multi-repo roots get a stray `.git/` dir created at the root *during plan computation* (which raw mode can then roam), and the "pure function / no writes" doc at `reconcile.rs:520` is falsified. Fix: `repo_root` or `tempfile::tempdir()`.

**HIGH-2 — upload-time conflict veto livelocks reclassified FF pushes with concurrent clocks** (`engine.rs:1626-1654`, tick at `:1544`): the tick only dominates when plan-time clocks were **Equal**. For **concurrent** clocks — the actual G5 field case and the merge-recovery case — the upload veto-skips (`skipped=true`), nothing errors, and every cycle replans the same Push forever. The feature silently fails on its own target scenario, and the "atomic per-repo winner" splits at execution time. A reclassified FF push must carry `merge(local,remote)+tick` (or an ancestry-justified force past the veto).

**MEDIUM-1 — objects-before-refs is ordering, not a barrier** (`reconcile.rs:1299`, `:1424-1468`, concurrent path `:1577`): a failed object pull still lets the later ref apply → ref points at a missing object; mirror on push publishes a ref whose objects never landed. Fix: skip ref-class actions for any repo with an object-class error this run, both directions.

**MEDIUM-2 (follow-up ticket ok)** — lock story overstated: acquire-failure proceeds unlocked (`reconcile.rs:738-746`), no staleness handling despite the Drop comment (`git_safety.rs:386-421`) so a SIGKILL leaks `tcfs.lock` permanently → warn+proceed unlocked forever; `refs/heads/*.lock` literal never globs (`git_safety.rs:37`).

**MEDIUM-3 — test gaps**: the RemoteAhead→Pull reclassifier arm (`reconcile.rs:1050-1060`) has zero coverage (device A converges via plain vclock RemoteNewer); the HIGH-2 veto case and DEFER path are untested. The `classify_fast_forward` docstring (`git_safety.rs:194-199`) wrongly claims objects are present at reclassification time (reclassify runs at plan time, before pulls).

Verified-good (briefly): ancestry direction + SHA sources correct both sides; fail-closed everywhere in the reclassifier; flag default-off, non-git plan order byte-identical; the same-second stat-bypass fix is sound (re-snapshot+re-hash at execute, `pin_mtime_to_cached_second` is a nice deterministic repro); #515 fence interplay clean.

Fix round in flight. Merge stays gated on: fixes + CI green + #515/#506 landed + harness-flip packet.

Jesssullivan:
## Fix round landed + adjudicated (11e33f7)

All four review findings addressed, verified by diff inspection + GitHub CI green:

- **HIGH-1**: ref-blob temp files now under `std::env::temp_dir()/tcfs-ff-{pid}-{seq}` — planning can no longer fabricate `.git` under any sync root; "pure function" doc corrected.
- **HIGH-2**: reclassified FF pushes now `merge(local, remote) + tick(device)` — the ancestry proof (local is a strict git descendant) justifies dominating a concurrent remote clock. Ordinary pushes keep the veto unchanged. The livelock-on-concurrent-clocks case is covered by a new test that uploads + converges.
- **MEDIUM-1**: objects-before-refs is now a real per-repo barrier (`git_ref_barrier_hit`), both sequential + concurrent paths; barred refs reported as `deferred_git_refs` (info, not errors), self-heal next cycle.
- **Doc/lock truth**: classify_fast_forward docstring corrected; lock staleness handling added (git_safety +204).

New tests: `git_ff_concurrent_clocks_push_uploads_and_converges`, `git_ff_remote_ahead_concurrent_clocks_reclassifies_to_pull`, `git_object_failure_bars_ref_actions_sequential`, `git_object_failure_bars_ref_pull_concurrent`, `git_ff_planning_fabricates_nothing_under_root`.

**Acceptance state**: fence #515 ✅ merged · harness #506 ✅ merged · CI ✅ green · harness-RED baseline banked (divergent per-file tie-break → `invalid sha1 pointer`; #513 closes it by never per-file-resolving `.git`: FF → atomic winner, divergent → zero writes) · **remaining: operator review**. Divergent keep-both (T10/T11) stays an honest open follow-up.

Jesssullivan:
# ADVERSARIAL FINAL-GATE REVIEW — PR #513 @ a12e430

## VERDICT: **BLOCKED** (one deterministic silent-clobber gap; one racy lost-pointer window)

Evidence base: full PR diff vs merge-base 2f3c046 read line-by-line; targeted test runs in isolated worktree (`git_ff_resolution`: 7/7 pass, 11.5s; `git_lock` unit tests: 3/3 pass); pre-fix commit archaeology for test-vacuity checks. Worktree removed after review; operator checkout untouched.

---

## BLOCKER-1 (deterministic, no race needed): non-head ref state rides the group-win + dominance with NO ancestry proof

The proof loop in `decide_repo_fast_forward` (reconcile.rs:1147+) examines **only** `.git/refs/heads/*` conflicts — `head_ref_for_git_path` (git_safety.rs:303) matches the `.git/refs/heads/` needle exclusively. But the apply loop (reconcile.rs:1083–1124) rewrites **every** `.git/*` Conflict in the repo group. So `packed-refs`, `refs/tags/*`, `refs/stash`, `refs/remotes/*` are force-pushed with the `GitFastForward` dominance ticket (reconcile.rs:1101) — or force-pulled — based on a proof about a *different file*.

Concrete kill chain (tags): both devices tag `v1` at different commits; loser pushes first (remote entry = theirs). Winner's `main` is genuinely FF-ahead → group wins LocalAhead → winner pushes its `v1` with dominance: `merge(local, remote_plan) + tick` (engine.rs:1633–1643). The execute-time guard **passes** — the remote entry for the tag hasn't moved since plan; it IS the loser's write, and dominating it is exactly what the code does. The dominated clock now strictly exceeds the loser's clock, so the loser's next reconcile sees plain **RemoteNewer and silently pulls the clobber down. No conflict is ever surfaced anywhere.** Same for `refs/stash` (both machines stashed — the exact bidirectional regime this feature targets; loser's stash pointer plus its `logs/refs/stash` reflog both ride the group) and worst for `packed-refs`: auto-gc packs **all** branch tips, so a winner's stale `packed-refs` rewinds branches the proof never examined; if the loser's tip lived only in packed-refs (loose ref pruned by gc), the pointer is gone modulo `fsck --lost-found`. Conflicts persist across cycles, so a week-old packed-refs conflict rides along the first time any head proves FF.

This violates the PR's own stated contract ("anything else … leaves the repo's conflicts untouched (fail-closed)") — divergence in ref-*valued* files outside `refs/heads/` is fail-**open**. This is the HIGH-2 trap realized: the "remote moved since plan" guard is correct for what it checks, but the dominance is granted to files the ancestry proof never covered.

**Required change (small, surgical):** in `decide_repo_fast_forward`, veto the repo (return None) when the conflict group contains any ref-valued path that is not a provable head — i.e. `packed-refs`, or `.git/refs/**` outside `refs/heads/` — OR extend the proof: parse both packed-refs (the remote blob is already downloadable via the `read_remote_ref_sha` machinery) and require every ref equal-or-FF in the winning direction; tags require equality. Index/logs can keep riding (workdir state, by design). A regression test: head FF-ahead + divergent `refs/tags/v1` in the same repo must stay Conflict.

## BLOCKER-2 (racy): proof earned at plan time, spent on execute-time state — no local-side re-verify in either direction

- **Pull direction (worse):** reclassified pulls are plain `PullReason::RemoteNewer` (reconcile.rs:1116) — indistinguishable at execute, **zero** execute-time guard (the HIGH-2 fix guarded only pushes). `download_file_with_device` overwrites via temp+rename unconditionally. Ref-class pulls sort after ALL object pulls, so the window is plan → end-of-object-wave — **minutes** on a large cycle. A `git commit` on the losing device in that window (the user is actively working — that's why clocks conflicted) gets its branch ref AND its reflog (`.git/logs/**` is in the group) overwritten → the fresh commit dangles with no pointer anywhere. This is silent local lost work that the pre-PR "conflict = zero writes" rule prevented.
- **Push direction:** upload re-hashes the *current* file and pushes it under the dominance earned by the plan-time SHA (engine.rs:1633; no check that the local ref still equals the proven SHA). A mid-window `git reset --hard`/amend publishes an unproven rewind with a dominating clock — peers then see RemoteNewer and pull it. This one partially self-heals (any device still ahead FF-pushes back), but the dominated clock also converts peers' rewind-pulls to silent.
- The claimed mitigations don't hold: `tcfs.lock` is TCFS-cooperative only — **git never checks it** (the comment at reconcile.rs:1408–1415 "stops a concurrent commit" overclaims); lock-acquire failure proceeds unlocked (reconcile.rs:792); a busy repo skips only the **lock**, not the repo's actions (reconcile.rs:775–784 — read the log message vs the control flow).

**Required change:** carry the plan-time local ref SHA in both reclassified directions (extend `GitFastForward` and add a pull-side marker) and byte-compare the local ref file immediately before dominating (push) or overwriting (pull); mismatch → defer, re-plan. Cheap, closes both directions.

## What I attacked and could NOT break (honest positives)

1. **HIGH-2 remote-side trap as posed:** guarded. Membership is plan-time but each push re-reads the index entry at upload (engine.rs:1570–1576) and only dominates if hash == plan pin (engine.rs:1636); remote movement → veto + re-plan. The residual read-to-publish race degrades to a *detected* conflict (the concurrent write's tick is never swallowed — merge is of the pinned manifest's clock), not silent loss. Partial group application is possible (some paths dominated, some vetoed) — "atomic per repo" is plan-time only; mixed remote state converges next cycle. Note only.
2. **Barrier:** keying is consistent (`repo_root_for_git_path` both sides, per-repo BTreeSet; no cross-repo over-barring; nested-repo roots distinct). Sequential top-of-loop check covers Push AND Pull (reconcile.rs:1426–1438); concurrent waves are strictly sequential (wave-1 spawns only after wave-0 joins complete) — no shared-set race. Object-class conflicts are content-addressed so veto-skip-without-Err on objects is unreachable in practice.
3. **Lock staleness:** PID-reuse fails SAFE (recycled live PID → never stolen — availability only); future-mtime/parse failure → fail-closed. Check-then-unlink race between two stealers can delete a just-recreated lock, but the consequence is bounded by the existing proceed-without-lock policy (nit).
4. **Test vacuity — all five are real:** concurrent-clocks test builds genuine concurrency via `tick_tracked_vclock` + asserts the remote manifest actually changed (fails on old veto-skip) + no-livelock replan + provenance (no-FF config yields Conflict); barrier tests use real failures (nonexistent manifest / missing local file) against refs that WOULD succeed, and the concurrent one genuinely routes to the concurrent path (dispatch predicate reconcile.rs:1364–1384: all-NewRemote, no encryption/progress); fabricates-nothing would catch the actual old bug (verified: pre-fix code created `local_root/.git/tcfs-ff-tmp` — the test's `!root/.git` assert fires). Nit: residue grep is name-coupled to "tcfs-ff".
5. **Merge state (#515/#506):** clean. Fence denies `.git/worktrees/**` at *collection* (blacklist.rs:285/315/342), so fenced paths never become actions and never enter FF groups; Taskfile changes are additive #506 harness tasks, no artifacts; full test file compiles+passes at a12e430.
6. **Regression sweep:** `GitFastForward` has exactly one constructor (reconcile.rs:1100) and one consumer (reconcile.rs:1477); all other `upload_file_with_device_with_state` callers pass `None`. Non-raw/daemon users: `git_ff_resolution` gated on `allows_git_dirs() && mode=="raw"` in both production callers (daemon.rs:1106, main.rs:6454); derived `Default` is off. The switch of ALL planned pushes to bypass the stat quick-check is an intended fix (plan classified by content hash), bytes identical.

## MEDIUM findings (fix with the blockers or ticket immediately)

- **M-3: `.git/tcfs.lock` roams.** No blacklist/collector exclusion exists (verified: zero hits in blacklist.rs/engine.rs). A crash-leaked lock, or a daemon scan racing a CLI execute, collects and syncs the lock to every peer — where a foreign PID probed against the local pid table can make it permanently unstealable → fleet-wide proceed-without-lock + the lock file itself joining `.git` conflict groups. Add `tcfs.lock` (and arguably git's own `*.lock`) to the deny set.
- **M-4: submodule gitdirs unprotected.** `.git/modules/dep/{objects,refs}/**` matches neither `is_git_object_class_path` nor `is_git_ref_class_path` (reconcile.rs:710/721 needles are `.git/objects/`/`.git/refs/`) → no objects-before-refs ordering, no barrier, no failure marking for raw-roamed submodules.
- Environmental: `TCFS_ASSUME_FRESH_PREFIX=1` (still live on neo per fleet state) skips the entire conflict/dominance block (engine.rs:1564) — FF pushes would publish with no guard at all. Not a PR defect; a deployment precondition worth stating in the PR.

## Bottom line

The fix round genuinely closed HIGH-1, the remote-side half of HIGH-2, and MEDIUM-1 — the engineering there is sound and well-tested. But the dominance mechanism now silently launders divergence in every ref-valued file the ancestry proof doesn't cover (tags/stash/packed-refs — deterministic, no race), and neither direction re-verifies local state at execute (racy ref+reflog clobber of never-pushed commits). Both violate the feature's own "divergent stays Conflict, zero writes" contract; both fixes are small (veto-or-extend the proof; pin+verify local SHA). BLOCKED until those two land; M-3/M-4 can be same-PR or immediate follow-ups.

Jesssullivan:
## Adjudication packet landed (wave-1 H1, ledger item 10) — [PR #520](https://github.com/Jesssullivan/tummycrypt/pull/520)

Dated evidence + one-page `VERDICT.md` for the operator rubber-stamp:

- **Harness RED baseline re-run on post-#514 `main` (2f3c046)**: flip-flop invariant holds; midwrite gate recorded; conflict row **CORRUPTION-RISK CONFIRMED** (`invalid sha1 pointer`) — G5-git-5 stays **expected-red**.
- **`git_ff_resolution` at this PR's head (a12e430): 7/7 pass.**
- Both blockers from the adversarial final-gate review above were **independently re-verified in source** before adoption (B1: proof loop covers only `refs/heads/*` while the apply loop rewrites every `.git/*` conflict with dominance; B2: pull side reclassifies to plain `RemoteNewer` with zero execute-time guard, push side re-checks remote only). M-3 (`tcfs.lock` absent from blacklist) also re-verified.

**VERDICT: DO NOT MERGE YET.** Unblock path (in `VERDICT.md`): land B1+B2 fixes (+ head-FF-with-divergent-tag regression test, mid-window local-commit defer test), M-3/M-4 same-PR or ticketed, re-run suite + harness, then operator merge. The merge stays operator-gated.

[wave1-2026-07-01]

Jesssullivan:
# THIRD adversarial gate — PR #513 @ 420446d

## VERDICT: **BLOCKED** — one silent-clobber gap survives the fix (submodule ref-valued paths bypass the BLOCKER-1 veto). BLOCKER-1 (top-level) and BLOCKER-2 (both directions) are genuinely, well closed. M-3/M-4 solid, tests non-vacuous, no regressions in the prior verified-goods.

Evidence base: fix-round diff `a12e430..420446d` read line-by-line; classifier logic replicated and executed standalone; `git_ff_resolution` 10/10 pass (11.0s) + `tcfs-sync --lib` 223/223 pass in an isolated detached worktree at 420446d. Worktree removed; operator checkout untouched.

---

## BLOCKER-3 (deterministic, no race): submodule ref-valued paths ride the FF group with no ancestry proof

The BLOCKER-1 veto `is_ref_valued_non_head_path` (reconcile.rs:794), called in `decide_repo_fast_forward` (reconcile.rs:1287), matches only `.git/packed-refs` (via `ends_with`) and `.git/refs/` (via `find`). It does **not** match a submodule's real gitdir at `.git/modules/<name>/…`. The FF prover `head_ref_for_git_path` (git_safety.rs:303) likewise matches only `.git/refs/heads/`, so submodule head refs are invisible to it too. Yet `repo_root_for_git_path` (git_safety.rs:275) splits on the **first** `.git` component — so every `.git/modules/<name>/**` conflict groups under the **outer** superproject root and rides its FF decision.

Empirically confirmed (exact replicas of the shipped functions):

```
repo/.git/refs/tags/v1                 veto=true   head_ref=None   ← top-level: caught ✓
repo/.git/modules/dep/refs/tags/v1     veto=false  head_ref=None   ← submodule tag: NOT caught
repo/.git/modules/dep/packed-refs      veto=false  head_ref=None   ← submodule packed-refs: NOT caught
repo/.git/modules/dep/refs/stash       veto=false  head_ref=None   ← submodule stash: NOT caught
repo/.git/modules/dep/refs/heads/main  veto=false  head_ref=None   ← submodule branch: neither vetoed NOR proven
```

Ironically the M-4 fix explicitly pulled `.git/modules/<name>/**` into the FF ordering/barrier machinery (`is_git_modules_ref_class`, reconcile.rs:775) — formalizing that submodule refs flow through this path — while leaving the BLOCKER-1 safety veto blind to them. The comment claims submodule gitdirs are handled; the veto is not.

**Kill chain (submodule branch clobber — no race needed):** superproject roamed under `git_sync_mode=raw` + `sync_git_dirs=true`, with a submodule whose gitdir is `<super>/.git/modules/dep/`.
1. A and B both at superproject C0, submodule `dep` at S0.
2. A advances superproject `main` C0→C1, pushes (remote head clock gets a device-a tick). A also commits in `dep` → S1a, pushed.
3. B commits in `dep` → **S1b** (divergent from S1a; equal/concurrent clock on `.git/modules/dep/refs/heads/main`), and fast-forwards superproject `main` **past C1 → C2** (strict descendant, concurrent clock).
4. B reconciles. Group = { outer `.git/refs/heads/main` (provable LocalAhead FF), submodule `.git/modules/dep/refs/heads/main` (divergent) }. Veto loop: submodule ref → `false` → **no veto**. Proof loop: outer main proven FF, submodule ref → `head_ref=None` → skipped; `saw_ref=true`, direction=LocalAhead → returns `Some`. **All** group members, including the submodule branch ref, are rewritten to `Push::GitFastForward` and dominate the remote clock (engine dominance = `merge(local, remote) + tick`).
5. B publishes `dep`'s ref = S1b with a dominating clock. A's S1a is now strictly dominated → A's next reconcile sees plain RemoteNewer → **silently pulls S1b, clobbering the S1a pointer.** The submodule reflog `.git/modules/dep/logs/refs/heads/main` is in the same group and rides too, so A's S1a commit dangles with no pointer anywhere (recoverable only via `fsck --lost-found`). **Silent loss of committed submodule work.**

BLOCKER-2's pins don't help: the pins carried on every group action are the **outer** head-ref SHA; they re-verify the outer ref, not the submodule ref, so an unchanged outer head lets the submodule clobber through.

**Required fix (small, reuses existing helpers):** the veto is too specific. Fail closed on **any** ref-class path that is not a provable top-level head:

```rust
if is_git_ref_class_path(rel_path) && git_safety::head_ref_for_git_path(rel_path).is_none() {
    return None; // packed-refs, tags, stash, remotes, HEAD, AND all submodule ref-class paths
}
```

`is_git_ref_class_path` already returns true for submodule refs (M-4 test asserts it) and false for `.git/index`/`.git/logs/**` (which correctly keep riding). This also closes the detached-`.git/HEAD` variant noted below. Regression test: superproject head FF-ahead + divergent `.git/modules/dep/refs/heads/main` in the same group must stay Conflict.

---

## Confirmed CLOSED (attacked, could not break)

**BLOCKER-1 (top-level) — CLOSED.** `is_ref_valued_non_head_path` catches `packed-refs`, `refs/tags`, `refs/stash`, `refs/remotes`, `refs/notes` (empirically true for all). Veto loop scans the **whole** group before any decision (reconcile.rs:1285-1296) and returns `None` on first hit → entire group stays Conflict, zero writes. `git_ff_divergent_tag_vetoes_whole_group` is non-vacuous: builds a genuine equal-clock divergent tag alongside an FF-ahead head, asserts **both** head and tag stay Conflict and no ref push/pull — fails without the veto. `.git/index`/`.git/logs/**` correctly ride (head reflog follows its head; the stash **pointer** `refs/stash` is vetoed, which fires the whole-group veto in the normal stash case).

**BLOCKER-2 (both directions) — CLOSED.** `GitRefPin` carries the plan-time local SHA (reconcile.rs:44); `git_ff_pins_still_valid` (reconcile.rs:854) re-reads the live ref file + `parse_ref_sha` and defers on any mismatch or missing file (fail-closed). Guard runs immediately before apply for both Push and Pull (reconcile.rs:1594). Every group member (reflog + index included) carries the pins, so a moved head defers the **whole group** atomically — the reflog cannot overwrite while its ref defers. Packed-only ref state → loose read Errs → defers (safe over-defer). The concurrent pull path handles only `PullReason::NewRemote` (reconcile.rs:1846), so FF pulls never bypass the sequential guard. Both defer tests genuinely mutate the local ref (`commit`) between plan and execute and assert defer + remote-untouched (push) / fresh-commit-survives (pull) — fail without the pins.

**M-3 — solid.** `is_git_lockfile_path` (blacklist.rs) denies any `*.lock` inside a `.git/` dir, wired into both `check()` and `check_path_components()` before config gating (fail-closed regardless of `sync_git_dirs`/`raw`). Test covers tcfs.lock, index.lock, HEAD.lock, `refs/heads/main.lock`, packed-refs.lock, nested-submodule index.lock; `Cargo.lock`/`flake.lock` (outside `.git`) still roam.

**M-4 (ordering) — solid.** `is_git_modules_{object,ref}_class` correctly classify submodule internals for objects-before-refs ordering + barrier, including nested `modules/`; the two axes stay disjoint. (This is precisely why the veto gap above matters — submodule refs now demonstrably flow through the FF path.)

**No regressions in prior verified-goods:** reclassify still gated `git_ff_resolution && git_sync_mode=="raw"` (reconcile.rs:651); single push-dominance constructor (now reconcile.rs:1218) + the new parallel pull marker; non-git paths byte-identical; barrier keying via `repo_root_for_git_path` unchanged.

---

## Notes / follow-up tickets (not blocking)

- **Detached `.git/HEAD` clobber** (same class, minor): a detached, divergent `.git/HEAD` (raw commit SHA) is neither vetoed nor proven and would ride an FF from a *different* branch's ref in the same repo. The recommended veto formulation above closes it for free.
- **Stash reflog-only edge:** `.git/logs/refs/stash` rides by design; the whole-group veto normally fires via the `refs/stash` pointer, but if only the reflog stack below the top differs (coincidentally identical top-stash object — very unlikely), lower stash entries could be overwritten. Low probability; ticket.
- **Over-veto by design (T10/T11 keep-both):** any benign divergent non-head ref (e.g. a stale remote-tracking ref) now blocks the whole repo from converging even when safe. Correct fail-closed behavior; a smarter per-ref proof / keep-both is the future unlock.
- **Pin-defer while actively committing:** a repo defers each cycle the local head keeps moving. Correct (never clobber active work); converges once the user stops. Not a livelock.
- **Residual TOCTOU on the pin itself:** microsecond window between `git_ff_pins_still_valid` and the actual write (tcfs.lock is cooperative; git ignores it). Vastly smaller than the plan→execute window the fix closed; inherent to check-then-act. Note only.
- **Env precondition:** `TCFS_ASSUME_FRESH_PREFIX=1` (still live on neo per fleet state) bypasses the whole conflict/dominance block — FF pushes would publish with no guard. Deployment precondition, not a PR defect; worth stating in the PR.

**Bottom line:** the fix round is high quality — BLOCKER-1/2 and M-3/M-4 are correctly implemented and non-vacuously tested. But the FF dominance still launders divergence in **submodule** ref-valued files (tags/stash/packed-refs/branch-heads) with no proof — a deterministic, silent, committed-work-loss path in a supported raw-roam config, the identical class the prior round blocked on. One-line veto broadening (`is_git_ref_class_path && head_ref_for_git_path().is_none()`) closes it. BLOCKED until that lands + a submodule-divergence regression test.

PR #524, opened 2026-07-05
ci(nix): build+publish full aarch64-darwin closure (incl. tcfsd) off-neo — fixes the neo local-compile trap

Jesssullivan:
## Current correction — 2026-07-05

Do not treat this PR as evidence that PZM is hardware-blocked or that the external SSD is the current TCFS blocker. Follow-up investigation found the SSD/APFS path re-enumerated and the current remote-builder failure is the Determinate Nix `ssh-ng://` remote-store path (`nix-daemon --stdio`, `MonitorFdHup kevent add: Invalid argument`), with `ssh://`/serve-style transport and non-PZM CI/cache lanes as the durable direction. This PR is therefore a tactical off-neo cache/prebuild lane only, and its cancelled macOS check must be refreshed before any merge consideration.

---

## Why
On 2026-07-04, `nix-switch` on **macbook-neo** silently compiled tcfs **locally for ~a full day**. Current corrected root-cause framing:
1. The SSD/APFS path on **petting-zoo-mini** re-enumerated and verified clean; do not treat PZM hardware or the external SSD as the current TCFS blocker.
2. The observed remote-builder failure is specific to Determinate Nix 3.17.3 `ssh-ng://` remote-store startup on macOS (`nix-daemon --stdio`, `MonitorFdHup kevent add: Invalid argument`).
3. This PR is therefore only a tactical off-neo cache/prebuild lane. The durable builder direction is `ssh://`/serve-style transport or GF/remote-cache execution patterns, with neo prevented from compiling locally.

## What this does
- Builds the **full** aarch64-darwin closure (adds `tcfsd`, `tcfsd-app`, `tcfs-file-provider-staticlib`) on the GitHub **macos-14** runner — genuinely off-neo, off-PZM.
- Publishes to **FlakeHub Cache** (`cache.flakehub.com` — the canonical fleet cache neo trusts; lab `nix.nix`: *"Attic replaced by FlakeHub"*) **and** keeps the **Attic** push (the GF-fabric nix-cache in the `nix-cache` namespace) extended to the daemon set.
- Net: neo `nix-switch` **substitutes** the darwin closure instead of compiling. PZM becomes optional, not a SPOF.

## Notes for review
- FlakeHub Cache push needs this repo **enrolled in FlakeHub Cache** (`id-token: write` added). The step is `continue-on-error` so CI stays green before enrollment; the Attic path works today with the existing `ATTIC_TOKEN`.
- The Bazel REAPI CAS at `grpc://bazel-cache.nix-cache.svc.cluster.local:9092` is **not** a nix-closure target (gRPC, cluster-internal) — the GF-fabric nix cache is the Attic in the same `nix-cache` namespace, which this already targets.

Pairs with the lab **fail-loud + preflight** PR so a dead builder errors in seconds instead of burning neo.

Linear: TIN-1620.

### DIFF
--- a/.github/workflows/nix-ci.yml
+++ b/.github/workflows/nix-ci.yml
@@ -97,6 +97,10 @@ jobs:
     name: Build macOS aarch64
     runs-on: macos-14
     needs: flake-check
+    # id-token: write is required for FlakeHub Cache OIDC push (below).
+    permissions:
+      contents: read
+      id-token: write
     steps:
       - uses: actions/checkout@v5
 
@@ -106,6 +110,17 @@ jobs:
           extra_nix_config: |
             extra-substituters = https://nix-cache.tinyland.dev/main
             extra-trusted-public-keys = main:eaUydxuDu7xBoy5cCo3MdknYAkVyTIASQ7DGuwxa+XA=
+
+      # FlakeHub Cache is the canonical nix cache the fleet substitutes from
+      # (lab nix/home-manager/nix.nix: "Attic replaced by FlakeHub"; macbook-neo
+      # trusts cache.flakehub.com). This action sets up a post-build push hook so
+      # EVERY darwin store path built below is published to FlakeHub Cache — which
+      # is what lets neo `nix-switch` SUBSTITUTE the aarch64-darwin closure
+      # (incl. tcfsd) instead of ever compiling locally. Requires this repo to be
+      # enrolled in FlakeHub Cache; no-ops with a warning if not yet enrolled.
+      - name: Set up FlakeHub Cache
+        uses: DeterminateSystems/flakehub-cache-action@main
+        continue-on-error: true
 
       - name: Configure Attic
         run: |
@@ -119,6 +134,12 @@ jobs:
             echo "ATTIC_CONFIGURED=false" >> "$GITHUB_ENV"
           fi
 
+      # NOTE: the aarch64-darwin DAEMON closure (tcfsd/tcfsd-app/staticlib) was
+      # previously NOT built here — only cli/tui/mcp — so macbook-neo could not
+      # substitute a new tcfsd and silently fell back to a LOCAL compile
+      # (2026-07-04 incident: a full day of neo compilation). Build the full set
+      # here on the GitHub macos-14 runner (off-neo, off-PZM) so the whole
+      # darwin closure lands in the caches and neo only ever substitutes.
       - name: Build tcfs-cli (no FUSE on macOS)
         run: nix build --accept-flake-config --fallback .#tcfs-cli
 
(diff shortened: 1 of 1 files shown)


Jesssullivan:
Observed CI root cause for the current `Build macOS aarch64` failure:

The job does not appear to be timing out inside the tcfs package builds. It reaches `DeterminateSystems/flakehub-cache-action@main`, downloads `magic-nix-cache-static-arm64-apple-darwin-0.2.0`, then that binary crashes on the GitHub macOS runner:

```text
dyld: Symbol not found: __ZNSt13exception_ptr31__from_native_exception_pointerEPv
Referenced from: /nix/store/...-magic-nix-cache-static-arm64-apple-darwin-0.2.0/bin/magic-nix-cache
Expected in: /usr/lib/libc++.1.dylib
```

After that, the action waits for the local cache service until the 6h job limit cancels the job. So this PR should not be treated as a mergeable/off-neo cache lane until the FlakeHub Cache action is pinned/fixed/disabled on macOS or the cache push is moved to a known-good path.

Current deploy path should remain the lab fail-closed remote-builder route (`ssh://...nix-store --serve` for PZM, not `ssh-ng://`) plus the lab flake bump, rather than rerunning this workflow as-is.

Jesssullivan:
2026-07-06 correction: this PR is still not merge-ready evidence for the off-neo deploy lane.

The current PZM blocker is no longer accurately summarized as only Determinate `ssh-ng://` / `MonitorFdHup`. Later lab probes show Finder/APFS can look healthy while SSH directory-health still times out under `/nix` and `/Volumes/TinylandSSD/tinyland`, with System Policy/FDA denials involving shell and Nix-store paths. PZM is therefore not accepted as a TCFS remote builder until lab passes directory-health, Nix execution-context probes, denial-log check, and `just nix-remote-builder-verify --live --strict`.

This PR remains useful as a tactical cache/prebuild idea, but its macOS aarch64 check timed out/cancelled and it should not be treated as the deploy fix. The current no-local-neo-build truth is captured in tummycrypt #537 plus lab #709/#723.

Jesssullivan:
Diagnosis of the two cancelled `Build macOS aarch64` runs: both died at exactly 6h00m (the hosted-runner ceiling) **inside "Set up FlakeHub Cache"** with every build step still `skipped` — the flakehub-cache-action hangs (rather than fails) when the repo isn't enrolled, and `continue-on-error` never fires on a hang. Pushed 2e22e36: `timeout-minutes: 10` on that step (degrades the hang to the warning the workflow comment always assumed) + `timeout-minutes: 300` on the job. The re-run should now reach the actual nix builds for the first time. Separately worth an operator look: enrolling the repo in FlakeHub Cache would make the step do its real job — it is the substitution path that lets neo nix-switch without compiling.

PR #529, opened 2026-07-05
feat(sync): add repo git keep-both resolver

Jesssullivan:
## Summary

Adds the PR-3 repo-group `.git` keep-both resolver for divergent branch-head conflicts. The resolver parks the remote branch tip under `refs/tcfs/theirs/<device>/heads/<branch>` while preserving the local branch head, then clears the conflict group only after Git safety checks and state flush succeed.

Refs TIN-1620 / G5-git keep-both lane. Stacked after #528, now rebased on `main`.

## Changes

- Adds `tcfs_sync::conflict_git::resolve_repo_keep_both` with dry-run and execute modes.
- Routes `tcfs resolve <repo> --strategy keep-both` through a dry-run by default, with `--execute` required for mutation.
- Adds daemon RPC routing for internal `git_keep_both_*` resolutions.
- Adds bounded state-cache snapshots so ref rollback does not leave in-memory conflicts falsely cleared.
- Tightens Git safety around branch `.lock` files and untracked worktree dirt.
- Adds focused regression coverage for missing remote manifests, moved local refs, parked-ref overwrite refusal, rollback behavior, untracked worktrees, state snapshot restore, and nested branch locks.

## Test plan

- [x] `~/.cargo/bin/cargo fmt --all -- --check` passes locally
- [x] `git diff --check origin/main..HEAD` passes locally
- [x] Focused read-only adversarial re-review reported no blockers
- [ ] `task test` passes locally
- [ ] `task lint` passes locally
- [ ] Verified affected functionality manually

Local build/test/clippy were intentionally not run on neo. Remote GitHub Actions are the build/test authority for this lane.

## Checklist

- [x] Docs updated (if user-facing behavior changed)
- [ ] CHANGELOG.md updated (if applicable)
- [x] No secrets or credentials in diff

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -322,16 +322,19 @@ enum Commands {
         state: Option<PathBuf>,
     },
 
-    /// Resolve a sync conflict for a file
+    /// Resolve a sync conflict for a file or git repo group
     ///
     /// When two devices modify the same file without syncing, a conflict is
     /// detected. Use this command to pick a resolution strategy.
     Resolve {
-        /// Path to the conflicted file
+        /// Path to the conflicted file, or a git repo root for repo-group keep-both
         path: PathBuf,
         /// Resolution strategy: keep-local, keep-remote, keep-both, or defer
         #[arg(long, short = 's', value_parser = ["keep-local", "keep-remote", "keep-both", "defer"])]
         strategy: Option<String>,
+        /// Execute repo-group git keep-both. Without this flag, repo mode is a dry-run.
+        #[arg(long)]
+        execute: bool,
     },
 
     /// List recorded sync conflicts (read-only)
@@ -906,14 +909,18 @@ async fn main() -> Result<()> {
         } => cmd_rm(&config, &path, prefix.as_deref(), state.as_deref()).await,
         Commands::Trash { action } => cmd_trash(&config, action).await,
         Commands::MigratePrefix { dry_run } => cmd_migrate_prefix(&config, dry_run).await,
-        Commands::Resolve { path, strategy } => {
+        Commands::Resolve {
+            path,
+            strategy,
+            execute,
+        } => {
             #[cfg(unix)]
             {
-                cmd_resolve(&config, &path, strategy.as_deref()).await
+                cmd_resolve(&config, &path, strategy.as_deref(), execute).await
             }
             #[cfg(not(unix))]
             {
-                let _ = (path, strategy);
+                let _ = (path, strategy, execute);
                 anyhow::bail!(
                     "resolve command requires the daemon (not available on this platform)"
                 )
@@ -6686,10 +6693,30 @@ async fn cmd_resolve(
     config: &tcfs_core::config::TcfsConfig,
     path: &Path,
     strategy: Option<&str>,
+    execute: bool,
 ) -> Result<()> {
-    let resolution = match strategy {
-        Some(s) => s.replace('-', "_"),
-        None => {
+    let is_git_repo = path.join(".git").is_dir();
+    let requested = strategy.map(|s| s.replace('-', "_"));
+
+    let resolution = match (is_git_repo, requested.as_deref()) {
+        (true, None) | (true, Some("keep_both")) => {
+            if execute {
+                "git_keep_both_execute".to_string()
+            } else {
+                "git_keep_both_dry_run".to_string()
+            }
+        }
+        (true, Some(other)) => {
+            anyhow::bail!(
+                "repo-group conflict resolution for {} supports keep-both only, got {other}",
+                path.display()
+            );
+        }
+        (false, Some(_)) if execute => {
+            anyhow::bail!("--execute is only valid for repo-group git keep-both resolution");
+        }
+        (false, Some(s)) => s.to_string(),
+        (false, None) => {
             // Interactive mode: reuse the existing interactive resolver.
             //
             // keep-both PR-1: fetch the REAL recorded ConflictInfo for this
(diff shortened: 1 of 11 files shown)


Jesssullivan:
## Adversarial review — BLOCKED (2 blockers + test gap + design-fidelity)

Winner-side mechanics are genuinely careful (no winner-repo corruption path found; dry-run side-effect-free; RAII lock released on all paths incl. panic; fsck gates both sides and actually abort; zero-OID CAS for idempotency; NATS auto still defers all .git). But this is a live `.git` WRITE path and two holes let it be reached the wrong way:

**BLOCKING 1 — execute path reachable from MCP / any socket client (violates the ratified operator-deliberate/CLI-only invariant → agent-triggered committed-work loss).** `git_keep_both_dry_run`/`git_keep_both_execute` were added to the shared `ResolveConflict` RPC allowlist (`grpc.rs:1660-1661`) and dispatched at `grpc.rs:1688-1690` **before** the per-file `.git` fence (`grpc.rs:1702`), with no operator-provenance gate. MCP `resolve_conflict` forwards an arbitrary `resolution` string (`tcfs-mcp/src/server.rs:249`); auth-bypass grants admin to any local socket client. So an MCP agent calling `resolve_conflict(rel_path=<divergent repo>, resolution="git_keep_both_execute")` parks the peer's heads + ticks the local clock to dominate — making the agent's device the fleet "winner" with no operator in the loop. Since PR-4 (loser no-loss guard) is not in this PR, the loser's next reconcile sees the winner's ref as clock-dominant RemoteNewer and overwrites `refs/heads/*`; the loser's old head is unreferenced → lost after gc. Fix: reject `git_keep_both_*` in the MCP passthrough AND gate the RPC dispatch on operator provenance (not a client-supplied magic string).

**BLOCKING 2 — undo bundle written in-tree under `.git/`, which the design forbids, and it roams.** `conflict_git.rs:413-423` writes `git bundle --all` to `repo_root/.git/tcfs-undo/keep-both-<uuid>.bundle`. Design §2.3 step 5 / S6: state dir, never in-tree. blacklist fences only `*.lock` + `.git/worktrees/**`, so this full-history bundle roams as an ordinary `.git/**` file on every execute, fresh uuid each time (never GC'd) → unbounded disk growth on a disk-constrained fleet, defeating the machine-local escape-hatch purpose. No `--restore-undo` consumer wired → pure write-only cost. Fix: write to the state dir (not under any sync root); optionally deny `.git/tcfs-undo/**` as belt-and-suspenders.

**CHANGES-NEEDED 3 — zero end-to-end Execute coverage.** The only test touching `resolve_repo_keep_both` runs DryRun + asserts a veto. No test builds two divergent repos, runs `--execute`, asserts theirs-parked-at-correct-SHA + local head kept + `git fsck --full` clean + both HEADs reachable + no lost commits + idempotent re-run. Inadequate for a live-`.git`-write path (design named G5-git-9/-11/-13 as PR-3's evidence bar; the #506 harness is unchanged here).

**CHANGES-NEEDED 4 — veto over-breadth → verb likely inert.** `collect_repo_conflicts` (`conflict_git.rs:275-280`) bails "unparkable" on any conflicted `.git` path that isn't `.git/refs/heads/*`. But `.git/index`, `.git/logs/HEAD`, `.git/COMMIT_EDITMSG` are non-ref-class and almost always divergent between two devices that each committed — design steps 7/9 intend these **kept-local**, not vetoed. Net: fail-closed and safe, but refuses on essentially every genuinely divergent repo. Reconcile with the design (keep index/logs local) so the verb is actually usable.

Minor (non-blocking): partial-object crash between update-ref and post-fsck can leave a stray theirs-ref that bricks future resolves until manual cleanup (`conflict_git.rs:378-381,193`); N>2 device-name sanitization collision fails-closed but spuriously refuses (`park_ref_for:347-357`).

Fix round incoming. Merge gated on 1+2 fixed, a real Execute end-to-end test, and the index/logs veto reconciled.

Jesssullivan:
## Re-verification (fix round aa374fa) — APPROVE

Independently verified the security-critical fixes against `origin/feat/keep-both-pr3-repo-resolve @ aa374fa3`:

**BLOCKING 1 CLOSED — two independent layers exclude MCP/agent/auto from the write path:**
- MCP `server.rs:252` rejects any `git_keep_both*` resolution BEFORE `connect()` (never sends the RPC) and hardcodes `operator_cli: false`; its `ResolveConflictInput` cannot express the field.
- Daemon `grpc.rs:1728` refuses `git_keep_both_*` dispatch unless `operator_cli` (proto field 3) is true; only the CLI (`main.rs:6762`) sets it. A raw-socket client could set it, but that is by definition the operator (raw socket already grants admin in auth-bypass) — the MCP/agent/NATS threat is fully closed.
- Fence ordering is correct: the repo-group path targets the repo ROOT (no `.git` component → `repo_root_for_git_path` returns None → fence doesn't fire), while per-file `.git` paths still hit the fence.

**BLOCKING 2 CLOSED:** undo bundle now written under the state dir (`<state_dir>/keep-both-undo/<repo-hash>/`, never a sync root); belt-and-suspenders `.git/tcfs-undo/**` fail-closed deny in blacklist (`GitTcfsUndo`), so any legacy in-tree bundle can never roam.

**CN-3:** real end-to-end Execute test (two divergent repos → parks theirs at correct SHA + winner untouched + fsck clean + both reachable + idempotent), fail-before/pass-after.

**CN-4:** veto reconciled — branch-heads park, genuinely-unparkable ref-class paths veto, `index`/`logs/**`/`COMMIT_EDITMSG` ride kept-local (winner's version; the parked theirs-ref preserves the loser's commits, so index/reflog need not materialize). Verb is now usable on real divergent repos.

Winner-side non-destruction, dry-run purity, RAII lock, fsck-both-sides-abort all still hold. CI 10/10 green. Merging as keep-both PR-3 (the divergent-half resolver).

PR #545, opened 2026-07-09
fix(state): CLI/daemon state-path convergence + conflict-preserving boot purge (TIN-2657)

Jesssullivan:
## TIN-2657 — CLI/daemon state-path convergence + conflict-preserving boot purge

The daemon and CLI agreed on `state.json` only for the *config-default* path. A
`--state <…/state.db>` override (and the `--state`-less `tcfs resolve`) resolved
verbatim to `…/state.db`, so the CLI read a different file than the daemon —
orphaning writes and making conflicts unreachable to `tcfs resolve` (live:
honey **TIN-2658**, the `git-roam-tool-daemon` 6-path conflict returned 0 refs).

### Approach

| Alternative | Verdict |
|---|---|
| Honor `config.sync.state_db` verbatim (drop `.with_extension`) | **Rejected — regression + silent data loss.** `StateCache::open` always uses `serde_json`; `state.db` is absent on every existing install → boots to an empty cache, losing all synced-file records, vclocks, and the TIN-2658 conflict record. Re-opens the daemon-reads-`.db`/CLI-reads-`.json` split #58/#59 already fixed. |
| Rename config key `state_db` → `state_path` | **Rejected — huge blast radius, no convergence benefit.** The key and literal `.db` are load-bearing across every fleet config and a dozen scripts; renaming needs a fleet migration and still doesn't fix the override-skips-normalization seam. |
| **Normalize override + guarded absorb (chosen)** | CLI and daemon converge on one `.json` for *every* input; existing `state.json` untouched; TIN-2658 records stay where the daemon already reads them; zero fleet steps; config key and `.db` defaults keep working. |

Three behavioral changes:

- **R1 — one path rule** (`tcfs-cli` `resolve_state_path`): the `--state`/`TCFS_STATE_PATH`
  override is normalized with `expand_tilde(p).with_extension("json")`, exactly
  like the config default. `.with_extension("json")` is idempotent for `.json`
  inputs, so `--state …/state.db` → `…/state.json` (the daemon's file) while an
  explicit `…/foo.json` passes through unchanged. Daemon/worker were already
  canonical and are unchanged.
- **R2 — absorb-on-boot (guarded)** (`tcfsd`): at boot, *only if* the canonical
  `state.json` is absent **and** a distinct sibling `state.db` exists, seed
  `state.json` from `state.db` once (copy). When both exist (honey/neo reality)
  `.json` always wins — no migration, no merge, no size heuristic. Copy failure
  is warn-and-continue.
- **Purge guard** (`tcfs-sync` `purge_stale`): the boot purge now never drops an
  entry carrying an unresolved conflict (`state.conflict.is_some()`), whatever
  its prefix. Previously any restart silently deleted TIN-2658's `git-roam/*`
  conflict record (foreign prefix) before it could be resolved.

Plus doc/comment truthing: `config.rs` `state_db` doc-comment now describes the
`.json`-sibling reality; the stale `crates/tcfs-daemon/...` path in
`docs/ops/current-workstream-truth-2026-07-06.md` → `crates/tcfsd/src/daemon.rs`.

### Tests

- `resolve_state_path_override_db_equals_default` — **the regression guard**:
  a `--state …/state.db` override resolves to the same file as the default.
  Verified revert-proof (fails on the pre-fix `return p.to_path_buf()`).
- `resolve_state_path_json_override_idempotent_and_tilde_expands` — `.json`
  passes through unchanged; `~/…/state.db` expands + normalizes to `.json`.
- `cli_write_via_db_override_visible_to_daemon_json_read` and
  `daemon_write_json_visible_to_cli_db_override_read` — both round trips.
- `absorb_seeds_json_from_db_only_host`, `absorb_leaves_existing_json_untouched_when_both_exist`,
  `absorb_neither_present_starts_fresh` — the R2 migration branches.
- `purge_stale_preserves_unresolved_conflicts_across_prefixes` — a foreign-prefix
  unresolved conflict survives the boot purge while a foreign non-conflict entry
  is still dropped.

### Scope

Refs **TIN-2657** (this fix) and **TIN-2658** (the conflict this unblocks for
`tcfs resolve`). Explicit non-goals kept out: `StateBackend`/`RocksDbStateCache`
dead code, the `state_db` config-key name, fleet TOMLs/scripts, `grpc.rs`, and
resolver-strategy code. The reconcile per-root isolation → `tcfs resolve`
routing gap remains TIN-2658 proper, flagged not fixed here.

Adversarial review gate to follow before merge — do not merge on green alone.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -1003,18 +1003,9 @@ async fn build_operator(config: &tcfs_core::config::TcfsConfig) -> Result<openda
     .context("building storage operator")
 }
 
-/// Expand `~` in path to the user's home directory
-fn expand_tilde(path: &Path) -> PathBuf {
-    let s = path.to_string_lossy();
-    if let Some(rest) = s.strip_prefix("~/") {
-        let home = std::env::var("HOME")
-            .or_else(|_| std::env::var("USERPROFILE"))
-            .unwrap_or_default();
-        PathBuf::from(format!("{}/{}", home, rest))
-    } else {
-        path.to_path_buf()
-    }
-}
+// `~`-expansion lives in tcfs-core so the CLI and daemon expand identically
+// (TIN-2657 adversarial gate, Fix B).
+use tcfs_core::config::expand_tilde;
 
 /// Resolve the state cache path: CLI flag > config > default user data dir
 fn resolve_state_path(
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -120,7 +120,9 @@ enum Commands {
         /// Remote prefix in the bucket (default: derived from local path name)
         #[arg(long, short = 'p')]
         prefix: Option<String>,
-        /// Path to the sync state cache JSON file (overrides config)
+        /// Path to the sync state cache JSON file (overrides config).
+        /// `.db` paths are normalized to their `.json` sibling — the file the
+        /// daemon owns — so the CLI and daemon always act on the same cache.
         #[arg(long, env = "TCFS_STATE_PATH")]
         state: Option<PathBuf>,
     },
@@ -136,7 +138,9 @@ enum Commands {
         /// Remote prefix to look up chunks (default: derived from manifest path)
         #[arg(long, short = 'p')]
         prefix: Option<String>,
-        /// Path to the sync state cache JSON file (overrides config)
+        /// Path to the sync state cache JSON file (overrides config).
+        /// `.db` paths are normalized to their `.json` sibling — the file the
+        /// daemon owns — so the CLI and daemon always act on the same cache.
         #[arg(long, env = "TCFS_STATE_PATH")]
         state: Option<PathBuf>,
     },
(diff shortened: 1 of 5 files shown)


Jesssullivan:
**Adversarial gate verdict: FIX-FIRST** (4 lenses, 12 agents, per-finding refutation; 8 findings raised, blocker confirmed).

**Blocker (must fix before merge):** `absorb_legacy_state_db` (daemon.rs:2084-2097) does a non-atomic, unvalidated `fs::copy` directly onto the canonical `state.json`. Mid-copy failure (ENOSPC is documented fleet history) leaves a truncated canonical file that (a) permanently satisfies the `!exists()` guard so absorb never retries, (b) fails `StateCache::open` with no `.bak`, silently rerouting the daemon to an ephemeral /tmp cache each boot — strictly worse than main for a .db-only host. **Fix A:** copy to `state.json.tmp` → parse-validate as StateCache → `rename` (the flush() idiom at state.rs:491-496); on failure remove temp, keep .db intact, `warn!` not `info!`.

**Bundled (same function, one-liner):** **Fix B** — tilde-expand `config.sync.state_db` before `.with_extension` (SyncConfig::default carries a literal `~`; absorb would otherwise write `./~/…` relative to CWD, reintroducing the exact divergence this PR closes). Latent today (all fleet configs absolute) but cheap now.

**Riding as follow-ups (refutation-confirmed, not gating):** per-root reconcile state files remain invisible to `tcfs resolve` (pre-existing, tracked under TIN-2658 — grpc.rs deliberately unchanged here); absorb/CLI boot race (no-worse-than-main); extensionless `--state` silent rewrite (latent, no live caller).

Full lens reports in the session record. Merge = Fix A (+B) pushed + CI green.

PR #572, opened 2026-07-29
fix(ci): restore fail-closed GloriousFlywheel authority (TIN-2538)

Jesssullivan:
## Summary

Restores TCFS CI to a source-owned, fail-closed GloriousFlywheel-only authority on exact base `dfe8282a4d93af0ca1c495065ec82002352d4d86`. This PR remains draft, source-only, digestless, and runtime-inert.

The repository-local public-read composite replaces the private cross-owner GF action that this public user-owned repository cannot consume. It keeps productive work on literal `tinyland-nix` and `tinyland-dind`, with no hosted fallback or copied private GF source.

## Current exact head

- signed head: `3d4de77933f2234d491dce1be2ac9e46e6fb4b10`
- parent: `292de66b5ef6d13b4d5c7cbf1231fc2e99e9de15`
- base: `dfe8282a4d93af0ca1c495065ec82002352d4d86`
- state: draft; auto-merge off

Two additive root-runner repairs sit above the reviewed 22-path authority packet:

- TIN-3799 replaces a mode-bit write-denial fixture with a regular-file parent, making the post-flush failure deterministic for every uid while proving the original file was not removed.
- TIN-3800 permits only root-level OS compatibility aliases such as macOS `/var`, and rejects nested root-owned repo aliases before Git capture, bundle, or ref side effects.

## Source evidence

- independent current-source review: P0=0/P1=0
- authority contract: 23/23
- TIN-3799 focused integration proof: 1/1
- `tcfs-sync` library suite: 441/441
- Cargo fmt and strict `tcfs-sync` clippy pass
- exact local action SHA-256: `8110be069476b984518dc6ff793592993d8a6e4a24b705389a83238a3aa43ed8`
- policy topology/action/command digests recomputed exact
- Actionlint, Bash syntax, Ruff, and diff hygiene pass
- complete authority manifest is unchanged by both one-file repairs

## Natural infrastructure findings

Prior sanctioned runs proved the public front door and then exposed three separate blockers:

- TIN-3798: the GF-owned `tummycrypt-dind` image lacks Docker Compose v2.
- TIN-545: exact-head evidence is cross-derivation and cross-runner: `rust-docs-1.93.0-x86_64-unknown-linux-gnu.drv` was killed by signal 9 on Linux job `94980810817` and Windows source-only job `94980810822`, while `tcfs-tui-0.12.17.drv` was killed on Nix job `94980811119`; no evidence-bound complete GF fix or quarantine exists yet.
[...]

PR #565, opened 2026-07-23
feat(sync): bind registered-root reconcile control state (TIN-2864)

Jesssullivan:
## Summary

Publishes the clean TIN-2864 registered-root reconciliation lane for canonical CI and independent review.

- binds immutable planner policy and strict reconciliation inputs
- captures held local snapshots and remote observation resources
- composes immutable GitRaw/catalog source state
- closes the source-only catalog publication, mutation-journal, monotonic-fence, and authenticated cold-recovery contracts
- composes current `main` and the TIN-2538 GF-only CI-authority source checkpoint

## Exact source state

- signed and GitHub-verified head: `0a3a9ceceeab497af614e07e80e91780c7f3be47`
- head commit: `fix(ci): run protected proof on stacked PRs`
- branch: `codex/tin-2864-root-plan-20260719`
- draft PR: 34 commits, 44 files, +43,368/-692
- Neo, Sting, upstream, and live remote are clean and exact at tree `6a1f26d6c00e478cbd007895aa364664d374f9f8`

## CP21 storage-fence closure

Every authoritative mutable Publishing `HEAD`, Index, Reservation, Manifest, and committed `HEAD` write must pass through one retained backend fence.

Each request exact-binds the selected root context, storage/control authorities, writer epoch and writer-fence fingerprints, Pending revision/record, mutation-journal ID and count, ordinal, stage, key, conditional-write predicate, payload length, and BLAKE3. The backend contract must atomically reject a stale fence or apply the exact write at the storage authorization point. Advisory liveness checks, an in-process mutex, or a control read followed by raw OpenDAL I/O do not satisfy it.

Raw rereads are receipt validation only and cannot redeem backend rejection. Exact replay and failure retention cover Pending, Publishing `HEAD`, namespace mutations, committed `HEAD`, and Ready.

## CP22 authenticated cold reconstruction

A fresh process can now reconstruct one exact canonical `PublicationPending` epoch without reusing warm planner/executor objects or treating copied bytes as authority.

- persists and reconstructs the complete predecessor storage binding, including Version+ETag
- fingerprints the canonical full binding and rejects ETag-only or crossed baselines
- reacquires an opaque handle to the same durable Pending fence and contiguous receipt ledger
- preflights aggregate count, key, body, binding, page, root, and per-kind budgets before untrusted bulk reads or retention
- reconstructs the immutable archived predecessor closure, authoritative journal, successor payloads, and exact unchanged named objects
[...]

PR #576, opened 2026-07-30
fix(sync): self-heal out-of-band rewrites instead of latching self-conflicts (TIN-3277)

Jesssullivan:
## Current delivery hold — 2026-08-18

This is a dated design/source packet, not a landable exact-head authority. The current PR head is `6453eb7abfaa3294f97aa38f800194d99c2298de`: all four carried commits are unsigned. Its merge base is `e7bfc1f01c3298c5ed6ad25254f6cc67d2a59213`; protected main is now `dfe8282a4d93af0ca1c495065ec82002352d4d86`, so the branch is 4 ahead / 1 behind and its historical checks are not final acceptance evidence.

After #572 and #565 land, reconstruct this patch semantically in a fresh clean worktree on the exact then-current base, create signed non-force replacement history, obtain independent exact-tree review, and require fresh natural sanctioned GloriousFlywheel CI. Do not rerun the old head, deploy, touch live state, or treat the Sting test transcript below as current runtime authority. This PR remains Draft with auto-merge off.

**v2** — rewritten after both adversarial refuters rejected v1. Every must-fix is addressed below. Still DRAFT; the re-verify pass decides readiness.

## The defect

An out-of-band writer (home-manager secrets/dotfile materialization on neo, an editor bypassing the tcfs write hook) rewrites an enrolled file **without ticking the vclock**. The clock freezes at its last-synced value while content diverges, so `conflict.rs::compare_clocks` sees EQUAL clocks + differing blake3 and returns `Conflict`. The reconcile `Conflict` arm is record-only — it bumps `times_recorded` and flips status, it never pushes — so the path is *structurally unpushable*. Live on neo: 7 permanent conflicts (`secrets/**`, `dotfiles/tcfs/devices.json`) stuck 24+ days with `times_recorded` in the thousands.

## The fix

When the pair is provably a local-only rewrite, tick a **comparison-only clone** of the vclock before `compare_clocks`, so it classifies `LocalNewer` → `Push` and converges. `conflict.rs` is untouched (`compare_clocks` byte-identical), so keep-both / loser-guard / FF-veto surfaces are unchanged.

### v2 predicate (`self_rewrite_retick_applies`)

```
  !device_id.is_empty() && remote_device == device_id   // corroboration ONLY, not the safety boundary
  && local_hash != remote_hash                          // identical content is UpToDate anyway
  && !is_git_internal_path(rel_path)                    // .git keeps the fail-closed FF-veto / keep-both path (G5-git-13)
  && tracked.is_some()
  && tracked.blake3 != local_hash                       // evidence the LOCAL file moved since last sync
  && tracked.blake3 == remote_hash                      // ★ NEW, LOAD-BEARING: the remote is our own last-synced bytes
  && stored_ordering == Some(Ordering::Equal)            // ★ computed against the STORED clock, before any re-tick
```

### The safety proof is CONTENT identity, not device identity

This is the central v2 change. `tracked.blake3 == remote_hash` means **the remote copy is byte-identical to what this device last synced**. Overwriting it therefore *provably* cannot lose another writer's work — there is nothing on the remote that is not already in this device's own tracked baseline.

Any foreign write — from any device, by any mechanism — changes `remote_hash`, fails the clause, and falls through to the ordinary `compare_clocks` → `Conflict` path exactly as before this PR. Concretely this closes:

- **A peer publishing a spoofed `written_by: "neo"`.** `written_by` (`manifest.rs`) is an unauthenticated client-set JSON string; `validate_indexed_manifest_binding` binds manifest *bytes* to the index's `manifest_hash`, it does not sign the device claim, and the registry-signing key is HKDF-derived from the shared master key. v1 made that string an authorization input. v2 does not: to pass the predicate an attacker must publish content *equal to our tracked baseline*, i.e. leave the remote exactly as this device last left it, which is a no-op.
- **A second isolated state cache under one `device_id`** (the macOS FileProvider, a `tcfs push` against a different state dir — memory records this as live: "0.12.17 resolve blind to isolated caches"). Its push changes `remote_hash`; predicate fails; conflict recorded.
- **A colliding / cloned `device_id`** (re-imaged host reusing a hostname-derived id). Same mechanism — content, not identity, decides.
- **`dotfiles/tcfs/devices.json` pushed stale.** If this device's local registry copy missed a legitimate revocation or enrollment that landed remotely, the remote content is *not* our baseline, so no heal — a recorded conflict, as today.

`written_by == device_id` is retained as corroboration, and because it preserves the pre-existing behavior that a distinct-device equal-clock pair records a conflict (and makes legacy v1 manifests, empty `written_by`, fail closed).

### No silent loss / why no parked copy

`KeepBoth` and the `.git` undo-bundle machinery exist to preserve a **second writer's** bytes. Under the content-identity clause the displaced remote content *is* `tracked.blake3` — this device's own last-synced bytes, still described by its own state entry and, until GC, still addressable by the manifest key. A `.conflict-{device_id}` sidecar would duplicate our own history, and on `secrets/**` would multiply ciphertext copies of sensitive material for zero recovery benefit.

In place of a sidecar, every self-heal push now logs at `info`: `path`, `device`, **`displaced_manifest`** (the remote manifest key being replaced), `displaced_hash`, `local_hash`. That is sufficient to reconstruct what was replaced.

## Refuter must-fixes → disposition

| # | Finding | Disposition |
|---|---|---|
| C-MF1 | **Double-tick promotion.** v1's predicate read a clock the TIN-2584 block had already ticked, so a strictly-dominated (`Less`) self-pair became `Equal` → ticked again → `Greater` → `Push`. | **Fixed structurally.** `stored_ordering` is computed **once**, against the stored clock, before any tick; the two rules are joined by `else if`; the predicate takes `stored_ordering: Option<Ordering>` as a parameter so a caller *cannot* hand it a mutated clock. Regression test `tin3277_self_pair_dominated_clock_is_not_promoted_to_push` pins main's behavior (recorded `Conflict`). Independently, the refuter's exact trace also now fails the content clause (`tracked.blake3 != remote_hash` there). |
| C-MF2 | **Missing `tracked.blake3 == remote_hash`** — the justification's own unstated premise; isolated-second-cache clobber with no `ConflictInfo`, no keep-both, GC-eligible chunks. | **Fixed.** Added as the load-bearing clause; the safety argument is rewritten around it. Negative test `tin3277_remote_moved_since_baseline_still_records_conflict`. |
| C-MF3 | **Vetoed-push path** may rewrite the tracked baseline and loop forever; `upload.skipped` invisible. | **Investigated + pinned end-to-end.** For the healed shape the veto is **unreachable**: the engine re-derives its clock from the tracked state with one `local_edit_inferred` tick (`engine.rs:2205-2215`), reaching `{neo:2}` vs remote `{neo:1}` → `Greater`/`LocalNewer`. `tin3277_full_cycle_self_heal_advances_clock_and_clears_latch` proves it with a real `plan → execute → state` cycle asserting `pushed == 1`. If the remote *does* move between plan and execute, `tin3277_veto_between_plan_and_execute_records_conflict_and_settles` pins the whole chain: push skipped, `pushed == 0`, a **visible** `ConflictInfo` recorded (`status = Conflict`, surfaced by `tcfs conflicts` / D-Bus), baseline moved to the live local hash — which **disarms** the predicate, so the next cycle settles into the pre-fix recorded-conflict behavior rather than replanning forever. Fail-closed, not silent. Plus: `execute_plan` now logs skipped pushes at `info` (they were in neither `pushed` nor `errors`). |
| C-MF4 | **Tests mirror rather than pin**; (c) vacuous, (d) can't mutate state, (f) hand-builds the post-push world. | **Reworked.** Vacuous (c) replaced by the content-clause negative case; hand-built (f) replaced by the real full-cycle test, which also asserts the latched `ConflictInfo` is *cleared* and `status` leaves `Conflict` (so the latch setup is load-bearing). Added the dominated-clock regression and the veto cycle test. **Mutation-tested** — see below. |
| S-MF1 | **Spoofed/colliding device identity → silent clobber.** | **Closed by the content clause.** `written_by` demoted to corroboration; documented in code and above. |
| S-MF2 | **`devices.json` self-heal-pushable**, bypassing the registry's Ed25519 verification. | **Narrowed, not closed** (corrected at the v3 gate). The content clause closes the specific property *"no clobber of remote bytes this device has not seen at `dotfiles/tcfs/devices.json"*. It does **not** demonstrate the different, undemonstrated property that the canonical Ed25519-signed registry (`tcfs-secrets/src/device.rs`, untouched by this diff) stays correctly reflected. No hard carve-out added, deliberately — see below. |
| S-MF3 | **`KeepBoth` / undo-bundle safety net bypassed; no parked copy.** | **Documented as redundant** under content identity (the displaced bytes are our own baseline), + `info` audit log of the displaced manifest key — now honestly labelled **best-effort plan-time provenance** (fields renamed `displaced_manifest_at_plan` / `displaced_hash_at_plan`), because the commit happens later after the engine's own fresh remote read. A remote that moves in between is vetoed and logged at execute time instead. Corrected at the v3 gate. |
| S-MF4 | **Secrets SSOT now unconditionally local-wins** with a TIN-3278-unreliable evidence gate and age non-determinism. | **Acknowledged, narrowed, and gated** — see below. |

## Explicit decisions the refuters asked for

**No `secrets/**` or `devices.json` carve-out.** With content identity, the dangerous scenarios are already excluded: a stale local registry can only heal when the remote is byte-identical to our own baseline, i.e. when the remote holds *no* revocation or enrollment we are missing. A carve-out would instead leave exactly the 7 live stuck paths permanently unhealable, which is the ticket.

**Age non-determinism is real and intended here.** Every home-manager re-materialization produces new ciphertext for identical plaintext, so post-fix each activation *will* push. That is the heal, not a bug — but it does mean secrets churn becomes visible push traffic.

**LANDING GATE (not a followup):** run a real ciphertext parity check neo↔honey for the 7 stuck paths before this reaches a host. Size parity is a weak proxy; decrypt-and-compare or recipient-set comparison is the real check. Promoted from a silent followup at the refuters' insistence.

**TIN-3278 interaction.** State-cache key-namespace duplication (absolute vs relative keys) affects 2 of the 9 entries in this triage and is a separate PR. It can make `tracked` resolve to the wrong entry, which would make the predicate fail (no heal) rather than fire wrongly — the content clause fails closed on a mismatched baseline. Not fixed here.

**Multi-host honesty.** This heals the **self-pair only**. The out-of-band writer runs on every host, so the equal-clock divergence class is inherently multi-host: peers keep accumulating record-only conflicts on the same paths, and the actual structural defect — the `Conflict` arm that records but never acts — is untouched. Expect some conflict count to migrate to honey rather than drop fleet-wide. Remaining TIN-3277 scope for a later PR: route HM writes through tcfs, or make the `Conflict` arm actionable (`AutoResolver` already returns `KeepLocal` for equal device names and has no caller on this path).

**Symlink blind spot.** `compare_both_exist_symlink` has the same equal-clock hole and is untouched (neo's dotfiles are a nix/HM symlink farm). Deliberately out of scope to keep this narrow.

## Gate round (v3) — second-pass must-fixes

Two independent adversarial reviews of v2 both returned **approve-with-must-fix**. All five must-fixes are applied on this head.

| # | v2 must-fix | Disposition |
|---|---|---|
| A-MF1 | **Prove the fix is not a no-op on its own target population** — the re-tick lives only in `compare_both_exist`; `compare_both_exist_symlink` has no re-tick at all, and `compare_both_exist` routes to it *before* any new code runs. neo's dotfiles are a nix/HM symlink farm and the headline test uses a hand-created regular file, which cannot disambiguate. | **Answered with live read-only evidence: all 9 conflicted entries in neo's state cache are REGULAR files — zero symlinks**, so this comparator is the right one and the fix is not a no-op. Probed via `lstat` on the entry keys in `~/.local/share/tcfsd/state.json` (no writes, no daemon interaction, freeze respected). The 7 `times_recorded = 4696` paths: `secrets/api/{github_token,anthropic,gitlab_token,crates_io_token}.age`, `secrets/infrastructure/tailscale_auth_key.age`, `secrets/.manifest.toml`, `dotfiles/tcfs/devices.json` — plus `secrets/.audit.log`, which appears **twice** under two key namespaces (`/Users/jess/tcfs/secrets/.audit.log` and `/secrets/.audit.log`, different `blake3`) — the live TIN-3278 duplication, in the flesh. Symlink parity stays a follow-up, correctly scoped. |
| A-MF2 | **No degenerate-local-content guard** on the one push no tcfs-aware actor initiated: `tracked.blake3 != local_hash` cannot tell "HM re-materialized the secret" from "HM/agenix failed and left a 0-byte or truncated file". | **Fixed in code, not prose.** Two fail-closed clauses: refuse when `local_size == 0` while `tracked.size > 0`; and for `*.age` paths refuse bytes that no longer carry an age container header (binary **or** armored). Keyed off the `.age` **extension**, not a `secrets/` prefix — because `secrets/.manifest.toml` is plaintext TOML by design and is one of the 7 live stuck paths, so a prefix rule would have made this fix a permanent no-op there (exactly the A-MF1 failure mode). Verified against the live corpus: the 5 stuck `.age` files all carry the **binary** header, so the guard admits them; a binary-only guard would have been fine but the armored form is accepted too, proven by test. Residual limit stated in the doc comment: a truncation that preserves the header is not detectable from content alone. |
| A-MF3 | **The new skipped-push log asserts "remote moved since plan"**, but `upload.skipped` is also true for benign `UpToDate` dedup, `RemoteNewer`, and the fast path. | **Fixed.** The message is now keyed off the engine's own `upload.outcome` (`Conflict` → "execute-time conflict veto: remote moved since plan"; `RemoteNewer`; `UpToDate` → "no-op: remote already holds this content"; else "skipped without a conflict verdict") under a neutral event name, with the verdict still logged as a field. |
| B-MF1 | **Plan/execute tracked-state divergence.** Planning resolves `tracked` via the fuzzy `StateCache::get_by_rel_path` (suffix match over the whole map, HashMap-order tie-break); the push resolves its own baseline via the exact canonical key (`StateCache::get`), and only the exact entry is ever written back by `state.set`. Under TIN-3278 the safety proof can be evaluated against a different record than execute mutates, with no detection. "Fails closed" was asserted, not proven. | **Fixed by proving it instead of asserting it.** `reconcile()` now also resolves the exact-key entry and proves the two lookups return the **same record** (pointer equality into the one entry map); `self_rewrite_retick_applies` takes `tracked_is_exact` and refuses when it does not hold, with a `warn!` naming TIN-3278. No other classification decision reads the flag, so nothing else changes behavior. The full-cycle test — which drives the real `reconcile → execute_plan → state` path — still heals, which is the proof that pointer equality holds on the real code path and this guard is not itself a no-op. |
| B-MF2 | **Audit-log integrity under race**: the `displaced_*` fields are read at plan time but the overwrite happens later, so they are not provably what was displaced — and this log is S-MF3's sole replacement for a parked copy. | **Fixed by correcting the claim** (the reviewer's own second option). Fields renamed `displaced_manifest_at_plan` / `displaced_hash_at_plan`; the comment and the S-MF3 row now say best-effort plan-time provenance, and point at the execute-time skip log for the moved-remote case. Emitting from the commit path would mean plumbing this through `engine::upload_planned_push_with_device`'s own remote read — a cross-cutting change with more blast radius than the honesty fix it buys; called out as a follow-up rather than smuggled in. |

Also taken from the reviewers' non-blocking notes: the doc comment no longer claims the content clause **provably** prevents loss — it now states the precise strength (`validate_indexed_manifest_binding` binds manifest *bytes* to `manifest_hash`, not the declared `file_hash` to the chunk payload, so the guarantee is "no honest remote writer's content is displaced"), and the noted coverage gap — that no test pinned the `else if` **independently** of the content clause — is closed by `..._same_content_republish_dominated_clock_is_not_double_ticked` (mutation E below).

## Mutation testing — the tests pin, they do not mirror

Run on sting against this exact tree:

- **Mutation A** — delete the `tracked.blake3 != remote_hash` clause → `tin3277_remote_moved_since_baseline_still_records_conflict` **FAILS** with `got Ok(Push { reason: LocalNewer })`. The content clause is load-bearing and the test catches its removal.
- **Mutation B** — A, plus restore v1's sequential structure and pass the live (already-ticked) clock ordering → **two** tests fail, including `tin3277_self_pair_dominated_clock_is_not_promoted_to_push` with `got Ok(Push { reason: LocalNewer })`. This is a verbatim reproduction of C-MF1's trace, caught by the new regression test.
- **Mutation C** — delete the `tracked_is_exact` clause → `tin3277_inexact_tracked_lookup_declines_self_heal` **FAILS** with `got Ok(Push { reason: LocalNewer })`.
- **Mutation D** — delete both degenerate-content clauses → **two** tests fail: `..._emptied_local_rewrite_is_not_pushed` and `..._non_ciphertext_local_age_rewrite_is_not_pushed`, both with `got Ok(Push { reason: LocalNewer })`.
- **Mutation E** — restore v1's sequential structure and live-clock ordering while leaving the content clause **intact** → `..._same_content_republish_dominated_clock_is_not_double_ticked` **FAILS** with `got Ok(Push { reason: LocalNewer })`. This is the test that pins the `else if` on its own; mutation B needed two edits to reach the same class of defect.
- Unmutated: all 13 pass.

## Tests (13)

| Test | Pins |
|---|---|
| `..._self_rewrite_equal_clock_pushes_instead_of_conflicting` | the heal itself |
| `..._distinct_device_equal_clock_still_records_conflict` | two-honest-device narrowing |
| `..._remote_moved_since_baseline_still_records_conflict` | ★ content-identity negative case |
| `..._self_pair_dominated_clock_is_not_promoted_to_push` | ★ double-tick regression (C-MF1) |
| `..._deferred_push_does_not_accumulate_ticks` | no per-cycle clock growth |
| `..._git_internal_self_pair_stays_conflict` | `.git` carve-out (G5-git-13) |
| `..._full_cycle_self_heal_advances_clock_and_clears_latch` | ★ real plan→execute→state: `pushed == 1`, clock `{neo:1}`→`{neo:2}`, latch cleared, next cycle no-op |
| `..._veto_between_plan_and_execute_records_conflict_and_settles` | ★ real cycle: veto → visible conflict → predicate disarmed, no loop |
| `..._same_content_republish_dominated_clock_is_not_double_ticked` | ★ v3: the `else if` **independently** of the content clause (mutation E) |
| `..._emptied_local_rewrite_is_not_pushed` | ★ v3: an emptied out-of-band rewrite never publishes |
| `..._non_ciphertext_local_age_rewrite_is_not_pushed` | ★ v3: plaintext over intact `*.age` never publishes — **plus** a `secrets/.manifest.toml` control proving the guard is an extension rule, not a prefix rule |
| `..._inexact_tracked_lookup_declines_self_heal` | ★ v3: baseline that may not be the record the push overwrites → fail closed |
| `..._age_header_guard_accepts_both_container_forms` | ★ v3: binary **and** armored age headers admitted; missing file / short read fail closed |

Every `*.age` test now uses real age-container bytes, so each verdict still turns on the clause it is testing rather than on an incidental header failure.

## sting results (verbatim)

All on sting (never on neo), scratch worktree `/tmp/tin3277-wt3`, removed afterwards.

```
$ cargo test -p tcfs-sync --lib tin3277
running 13 tests
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 436 filtered out; finished in 0.05s

$ cargo test -p tcfs-sync
test result: ok. 449 passed; 0 failed        # lib
... 13 integration suites, every one `test result: ok`, 0 failed anywhere

$ cargo fmt --all -- --check
FMT_CLEAN

$ cargo test -p tcfs-sync --lib tin3277
running 8 tests
test reconcile::tests::tin3277_self_rewrite_equal_clock_pushes_instead_of_conflicting ... ok
test reconcile::tests::tin3277_distinct_device_equal_clock_still_records_conflict ... ok
test reconcile::tests::tin3277_self_pair_dominated_clock_is_not_promoted_to_push ... ok
test reconcile::tests::tin3277_remote_moved_since_baseline_still_records_conflict ... ok
test reconcile::tests::tin3277_git_internal_self_pair_stays_conflict ... ok
test reconcile::tests::tin3277_deferred_push_does_not_accumulate_ticks ... ok
test reconcile::tests::tin3277_veto_between_plan_and_execute_records_conflict_and_settles ... ok
test reconcile::tests::tin3277_full_cycle_self_heal_advances_clock_and_clears_latch ... ok
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 436 filtered out; finished in 0.05s

$ cargo test -p tcfs-sync          # full crate
lib:                            444 passed; 0 failed   (13.09s)
auto_pull_logic_test:             6 passed; 0 failed
e2e_conflict_resolution:         11 passed; 0 failed
e2e_delete_scenarios:             6 passed; 0 failed
e2e_directory_ops:                6 passed; 0 failed
e2e_edge_cases:                   8 passed; 0 failed
e2e_file_state_transitions:       7 passed; 0 failed
e2e_placeholder_lifecycle:        4 passed; 0 failed
e2e_state_transition_chains:      6 passed; 0 failed
e2e_two_device_sync:              4 passed; 0 failed
state_cache_reload_test:          3 passed; 0 failed
symlink_handling_test:            8 passed; 0 failed
two_device_sync_test:             3 passed; 0 failed
Doc-tests:                        0 passed; 0 failed

$ cargo clippy -p tcfs-sync --all-targets -- -D warnings
Finished `dev` profile [unoptimized + debuginfo] target(s) in 52.84s     # no warnings
```

Scratch worktrees `/tmp/tin3277-wt2` (v2) and `/tmp/tin3277-wt3` (v3) on sting removed afterwards.

## Scope / collisions

One file: `crates/tcfs-sync/src/reconcile.rs`. `conflict.rs`, `engine.rs`, `state.rs` unchanged. Source-only under the deploy freeze — no config, no CHANGELOG, no deploys.

- **#565** touches `reconcile.rs` only to rename `validate_indexed_symlink_target` → `validate_restored_symlink_target_for_physical_path` at 5 symlink call sites, and adds `registered_reconcile.rs`. No functional overlap with the regular-file clock logic; mechanical rebase offset only.
- **#567 / #568** touch `registered_remote_catalog` / `publication.rs` / `tcfs-core/config.rs`. No overlap.

### DIFF
--- a/crates/tcfs-sync/src/reconcile.rs
+++ b/crates/tcfs-sync/src/reconcile.rs
@@ -1611,6 +1611,19 @@ pub async fn reconcile(
         let local = local_files.get(rel_path);
         let remote = remote_index.get(rel_path);
         let tracked = state.get_by_rel_path(rel_path).map(|(_, s)| s);
+        // TIN-3277/TIN-3278: `get_by_rel_path` is a fuzzy suffix match over the
+        // whole entry map, while the push path re-resolves its own baseline with
+        // the exact canonical key (`StateCache::get`) — and only that exact entry
+        // is ever written back. When duplicate keys exist for one file (live on
+        // neo today) the two lookups can disagree, and which one the fuzzy matcher
+        // returns is HashMap-iteration-order luck. Prove they are the SAME record
+        // (pointer equality into the one entry map) so the self-heal predicate can
+        // fail closed rather than reason about a record the push would not touch.
+        // No other classification decision consumes this flag.
+        let tracked_is_exact = match (tracked, local.and_then(|path| state.get(path.as_path()))) {
+            (Some(fuzzy), Some(exact)) => std::ptr::eq(fuzzy, exact),
+            _ => false,
+        };
 
         let action = classify_path(
             rel_path,
@@ -1621,6 +1634,7 @@ pub async fn reconcile(
             remote_prefix,
             device_id,
             config,
+            tracked_is_exact,
         )
         .await
         .with_context(|| format!("classifying reconciliation path {rel_path:?}"))?;
@@ -3158,6 +3172,7 @@ async fn classify_path(
     remote_prefix: &str,
     device_id: &str,
     config: &ReconcileConfig,
+    tracked_is_exact: bool,
 ) -> Result<ReconcileAction> {
     let action = match (local, remote, tracked) {
         // New local file — not on remote, not previously synced
@@ -3267,6 +3282,7 @@ async fn classify_path(
                 op,
                 remote_prefix,
                 device_id,
+                tracked_is_exact,
             )
             .await;
         }
@@ -3285,6 +3301,11 @@ async fn classify_path(
 }
 
 /// Compare when both local and remote exist — uses vector clocks.
+///
+/// `tracked_is_exact` says whether `tracked` is the state record that the push
+/// path would also resolve (exact canonical key) — see
+/// `self_rewrite_retick_applies`, the only decision here that depends on it.
+#[allow(clippy::too_many_arguments)]
 async fn compare_both_exist(
     rel_path: &str,
     local_path: &Path,
@@ -3293,6 +3314,7 @@ async fn compare_both_exist(
     op: &Operator,
     remote_prefix: &str,
     device_id: &str,
+    tracked_is_exact: bool,
 ) -> Result<ReconcileAction> {
     // Symlinks are first-class entries: they must NOT be dereferenced and hashed
     // like regular files (that would hash the *target's* content and then fail to
@@ -3407,37 +3429,64 @@ async fn compare_both_exist(
     // here would let a `Less` pair be re-ticked into `Equal` and then promoted
     // to `Greater`/`LocalNewer`, defeating the narrowing comment above.
     let stored_ordering = local_vclock.partial_cmp_vc(&remote_manifest.vclock);
+    // TIN-3277 local-side evidence, read from the same `symlink_metadata` call
+    // that chose this comparator, before any tick.
+    let local_size = local_metadata.len();
+    let local_content_plausible = local_age_ciphertext_intact(rel_path, local_path);
+    let self_rewrite_shape = |tracked_is_exact: bool| {
+        self_rewrite_retick_applies(
+            rel_path,
+            tracked,
+            tracked_is_exact,
+            &local_hash,
+            local_size,
+            local_content_plausible,
+            &remote_manifest.file_hash,
+            stored_ordering,
+            device_id,
+            remote_device,
+        )
+    };
     if tracked.is_some_and(|tracked_state| tracked_state.blake3 != local_hash)
         && stored_ordering == Some(std::cmp::Ordering::Less)
     {
         local_vclock.tick(device_id);
-    } else if self_rewrite_retick_applies(
-        rel_path,
-        tracked,
-        &local_hash,
-        &remote_manifest.file_hash,
-        stored_ordering,
-        device_id,
-        remote_device,
-    ) {
+    } else if self_rewrite_shape(tracked_is_exact) {
         // TIN-3277: an equal-clock self-pair whose remote content is byte-identical
         // to our own tracked baseline is proof of a local out-of-band rewrite, not a
         // concurrent-writer conflict. See `self_rewrite_retick_applies`.
         //
-        // Audit trail: log the remote manifest key we are about to displace. The
+        // Audit trail: log the remote manifest key we intend to displace. The
         // displaced bytes are `tracked.blake3` — this device's own last-synced
         // content — so this line is sufficient to reconstruct what was replaced
-        // without a parked copy.
+        // without a parked copy. It is emitted at PLAN time and the field names say
+        // so: the engine re-reads the remote before committing, so these values are
+        // the remote as observed during classification (best-effort provenance), not
+        // a post-hoc record of what the push actually overwrote. A remote that moved
+        // in between is vetoed at execute time and logged there instead.
         info!(
             path = %rel_path,
             device = %device_id,
-            displaced_manifest = %manifest_path,
-            displaced_hash = %remote_manifest.file_hash,
+            displaced_manifest_at_plan = %manifest_path,
+            displaced_hash_at_plan = %remote_manifest.file_hash,
             local_hash = %local_hash,
             "TIN-3277 self-heal: equal-clock self-pair, remote content == our tracked \
-             baseline — re-ticking as a local out-of-band rewrite and pushing"
+             baseline — re-ticking as a local out-of-band rewrite and planning a push"
         );
         local_vclock.tick(device_id);
+    } else if !tracked_is_exact && self_rewrite_shape(true) {
+        // Everything about the shape says "local out-of-band rewrite", but the
+        // baseline we would have proved it against is not the state record the push
+        // path would overwrite (TIN-3278 key duplication). Decline and let the
+        // ordinary Conflict arm record it — visible, and no push justified by a
+        // record that may describe a different file.
+        warn!(
+            path = %rel_path,
+            device = %device_id,
+            "TIN-3277 self-heal declined: plan-time tracked entry is not the exact-key \
+             record the push path would overwrite (TIN-3278 state-cache key duplication) \
+             — recording a conflict instead"
+        );
     }
 
     let outcome = compare_clocks(
@@ -3473,13 +3522,23 @@ async fn compare_both_exist(
 ///
 /// The safety proof is CONTENT identity, not device identity. The load-bearing
 /// clause is `tracked.blake3 == remote_hash`: the remote copy is byte-identical
-/// to what THIS device last synced. Overwriting it therefore provably cannot
-/// lose another writer's work — there is nothing on the remote that is not
-/// already in this device's own tracked baseline. Any foreign write, from any
-/// device, by any mechanism, changes `remote_hash`, fails this clause, and falls
-/// through to the ordinary `compare_clocks` -> `Conflict` path exactly as before
-/// this fix. That is what makes the promotion safe without authenticating
-/// anything.
+/// to what THIS device last synced. Overwriting it therefore cannot lose another
+/// writer's work — there is nothing on the remote that is not already in this
+/// device's own tracked baseline. Any foreign write, from any device, by any
+/// mechanism, changes `remote_hash`, fails this clause, and falls through to the
+/// ordinary `compare_clocks` -> `Conflict` path exactly as before this fix. That
+/// is what makes the promotion safe without authenticating anything.
+///
+/// Precision about the strength of that clause: it compares our baseline against
+/// the manifest's self-declared `file_hash` field.
+/// `engine::validate_indexed_manifest_binding` binds manifest BYTES to the
+/// index's `manifest_hash`; it does not prove the declared `file_hash` matches
+/// the chunk payload. A write-capable peer could therefore publish
+/// `file_hash == our baseline` over different chunks and satisfy this clause. So
+/// the guarantee is exactly "no HONEST remote writer's content is displaced",
+/// not an unconditional proof against a master-key-holding adversary — who can
+/// destroy data outright anyway, and whose forged payload the pull path's own
+/// content-hash verification would reject.
 ///
 /// `remote_device == device_id` is kept as CORROBORATION ONLY, deliberately not
 /// as the safety boundary. `written_by` (`manifest.rs`) is an unauthenticated,
@@ -3517,6 +3576,36 @@ async fn compare_both_exist(
 ///    (concurrent) is a real conflict. Taking the ordering as a parameter rather
 ///    than re-deriving it here makes it impossible for a caller to evaluate this
 ///    predicate against a clock a sibling rule already ticked.
+///  * `tracked_is_exact`: the baseline above must come from the SAME state
+///    record the push path will read and overwrite. Planning resolves `tracked`
+///    with the fuzzy suffix matcher `StateCache::get_by_rel_path`, while the
+///    push resolves its own baseline with the exact canonical key
+///    (`engine::upload_planned_push_with_device` -> `StateCache::get`) — and only
+///    the exact-keyed entry is ever written back by `state.set`. Under TIN-3278's
+///    live key-namespace duplication (absolute vs relative keys for the same
+///    file) those two can be DIFFERENT records, and `get_by_rel_path` returns
+///    whichever one `HashMap` iteration reaches first. Evaluating a
+///    content-identity proof against a record the push will not touch is exactly
+///    the ambiguity this predicate must not paper over, so the caller proves
+///    identity of the two lookups (pointer equality) and we fail closed when it
+///    does not hold.
+///  * `local_size != 0` while `tracked.size > 0`: the local side is the one thing
+///    here that no tcfs-aware actor authored, and the predicate's only other
+///    local evidence (`tracked.blake3 != local_hash`) cannot tell "home-manager
+///    re-materialized the secret" from "home-manager/agenix failed and left an
+///    empty file". Publishing the second case would replace good remote content
+///    with nothing and make the previous manifest+chunks GC-eligible, having just
+///    removed the latched Conflict that used to protect the path. Fail closed.
+///  * `local_content_plausible`: for `*.age` paths, the live bytes must still
+///    carry a recognizable age header (binary `age-encryption.org/v1` or the
+///    armored `-----BEGIN AGE ENCRYPTED FILE-----`), so a plaintext error string
+///    or truncated-to-nothing materialization is never published over intact
+///    ciphertext. Deliberately keyed off the `.age` extension rather than a
+///    `secrets/` prefix: the live stuck set includes `secrets/.manifest.toml`,
+///    which is plaintext TOML by design, and a prefix rule would make this fix a
+///    permanent no-op there. Residual limit, stated honestly: a truncation that
+///    preserves the header is not detectable from content alone and is NOT
+///    covered by this clause.
 ///  * `local_hash != remote_hash`: identical content is `UpToDate` anyway.
 ///  * not `.git`-internal: `.git` divergence keeps the fail-closed
 ///    fast-forward veto / loser-guard keep-both path (G5-git-13), which decides
@@ -3543,10 +3632,14 @@ async fn compare_both_exist(
 /// `tracked.blake3 != local_hash` for that path: the entry reverts to the
 /// pre-fix recorded-conflict behavior (surfaced by `tcfs conflicts` / D-Bus)
 /// rather than looping. Fail-closed, not silent.
+#[allow(clippy::too_many_arguments)]
 fn self_rewrite_retick_applies(
     rel_path: &str,
     tracked: Option<&SyncState>,
+    tracked_is_exact: bool,
     local_hash: &str,
+    local_size: u64,
+    local_content_plausible: bool,
     remote_hash: &str,
     stored_ordering: Option<std::cmp::Ordering>,
     device_id: &str,
@@ -3561,6 +3654,12 @@ fn self_rewrite_retick_applies(
     if is_git_internal_path(rel_path) {
         return false;
     }
+    // The baseline we are about to reason about must be the record the push path
+    // will read and overwrite, not whichever duplicate the fuzzy plan-time
+    // matcher happened to return (TIN-3278).
+    if !tracked_is_exact {
+        return false;
+    }
     let Some(tracked_state) = tracked else {
         return false;
     };
@@ -3573,7 +3672,47 @@ fn self_rewrite_retick_applies(
     if tracked_state.blake3 != remote_hash {
         return false;
     }
+    // The local side is the only side no tcfs-aware actor authored: refuse to
+    // publish a degenerate (emptied / non-ciphertext) out-of-band rewrite over
+    // intact remote content.
+    if local_size == 0 && tracked_state.size > 0 {
+        return false;
+    }
+    if !local_content_plausible {
+        return false;
+    }
     stored_ordering == Some(std::cmp::Ordering::Equal)
+}
+
+/// Number of leading bytes needed to recognize either age container header.
+const AGE_HEADER_PROBE_LEN: usize = 34;
+
+/// Do the live local bytes still look like age ciphertext?
+///
+/// Only meaningful for `*.age` paths; every other path answers `true` (nothing to
+/// check). Used as a fail-closed guard on the TIN-3277 self-heal push, which is
+/// the one push in this file that no tcfs-aware writer initiated: an out-of-band
+/// materialization that failed and left a plaintext error message must not be
+/// published over the intact ciphertext currently on the remote.
+///
+/// Accepts BOTH age container forms — the binary header (`age-encryption.org/v1`,
+/// which is what neo's live enrolled `secrets/**.age` files carry) and the
+/// ASCII-armored header — so the guard cannot silently disable the fix for a
+/// legitimately armored corpus.
+fn local_age_ciphertext_intact(rel_path: &str, local_path: &Path) -> bool {
+    if !rel_path.to_ascii_lowercase().ends_with(".age") {
+        return true;
+    }
+    let Ok(file) = std::fs::File::open(local_path) else {
+        return false;
+    };
+    let mut head = Vec::with_capacity(AGE_HEADER_PROBE_LEN);
+    let mut probe = std::io::Read::take(file, AGE_HEADER_PROBE_LEN as u64);
+    if std::io::Read::read_to_end(&mut probe, &mut head).is_err() {
+        return false;
+    }
+    head.starts_with(b"age-encryption.org/v1")
+        || head.starts_with(b"-----BEGIN AGE ENCRYPTED FILE-----")
 }
 
 /// Causal metadata extracted from one fully bound remote manifest.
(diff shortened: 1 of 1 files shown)


Jesssullivan:
## v1 → v2 (`7932d2c` → `fe69adb`)

Both refuters rejected v1. Rework below; PR body rewritten in full. Still DRAFT.

**The one change that collapses most of it:** the predicate now requires **`tracked.blake3 == remote_hash`**. The safety proof moves from *device identity* (an unauthenticated `written_by` string) to **content identity**: the remote copy is byte-identical to what this device last synced, so overwriting it provably cannot lose a second writer's work. Any foreign write — any device, any mechanism, including a spoofed `written_by` — changes `remote_hash`, fails the predicate, and records a `Conflict` exactly as before. `written_by == device_id` survives as corroboration only.

### Per finding

| Finding | Change |
|---|---|
| **C-MF1** double-tick promotes dominated self-pair to Push | `stored_ordering` computed **once** against the STORED clock before any tick; the two rules joined by `else if`; the predicate takes `stored_ordering: Option<Ordering>` so a caller *cannot* pass a mutated clock. Regression test `tin3277_self_pair_dominated_clock_is_not_promoted_to_push` pins main's `Conflict`. |
| **C-MF2** missing `tracked.blake3 == remote_hash` → isolated-second-cache clobber | Clause added; it is now the load-bearing one. Negative test `tin3277_remote_moved_since_baseline_still_records_conflict`. |
| **C-MF3** vetoed push rewrites baseline / loops / invisible | Veto proven **unreachable** for the healed shape via a real `plan → execute → state` cycle (`pushed == 1`; the engine's own `local_edit_inferred` tick reaches `Greater` independently). If the remote *does* move mid-cycle, a second cycle test pins the outcome: `pushed == 0`, **visible** `ConflictInfo` + `status = Conflict`, baseline moves to the live local hash which **disarms** the predicate → settles into pre-fix recorded-conflict behavior, no loop. `execute_plan` now logs skipped pushes at `info` (previously in neither `pushed` nor `errors`). |
| **C-MF4** tests mirror, don't pin | Vacuous latched test replaced; hand-built post-push world replaced by the real cycle test (which now also asserts the latch is *cleared*). Two new negative/regression tests. **Mutation-tested**, below. |
| **S-MF1** spoofed/colliding device identity → silent clobber | Closed by the content clause; `written_by` explicitly demoted to corroboration in code and body. |
| **S-MF2** `devices.json` self-heal-pushable past its Ed25519 gate | Closed by the content clause: a stale local registry can only heal when the remote holds no revocation/enrollment we are missing. **No hard carve-out** — stated and argued explicitly in the body, since a carve-out would leave the 7 live stuck paths permanently unhealable. |
| **S-MF3** `KeepBoth` / undo-bundle bypassed, no parked copy | Documented as redundant under content identity (displaced bytes *are* our own tracked baseline; a sidecar on `secrets/**` would multiply ciphertext for zero recovery benefit) + every self-heal push now logs `displaced_manifest` at `info` for auditability. |
| **S-MF4** secrets SSOT local-wins, unreliable gate, age non-determinism | Acknowledged head-on in the body: every HM re-materialization *will* push (intended). The neo↔honey ciphertext parity check is promoted from a followup to a **LANDING GATE**. TIN-3278 interaction documented — a mis-keyed `tracked` makes the predicate fail closed (no heal), not fire wrongly. |
| **Multi-host note** | Body now states plainly that this heals the self-pair only; peer-side accumulation and the record-only `Conflict` arm remain open TIN-3277 scope. No overclaim. |

### Mutation testing (the tests pin, they do not mirror)

- Drop the content clause → `tin3277_remote_moved_since_baseline_still_records_conflict` **FAILS**: `got Ok(Push { reason: LocalNewer })`.
- Drop it **and** restore v1's sequential structure / live-clock ordering → **two** failures, including `tin3277_self_pair_dominated_clock_is_not_promoted_to_push`: `got Ok(Push { reason: LocalNewer })` — a verbatim reproduction of C-MF1's trace.
- Unmutated: 8/8 pass.

### sting gate

`cargo fmt --all -- --check` clean · `cargo test -p tcfs-sync` = **444 lib + 72 integration passed, 0 failed** · `cargo clippy -p tcfs-sync --all-targets -- -D warnings` clean. Verbatim output in the PR body.

Jesssullivan:
## Adversarial gate: v1 reject → v2 approve-with-must-fix ×2 → fixed at the gate

**Round 1: reject.** Two independent refutations found a real correctness defect (the predicate read a clock the TIN-2584 block had already ticked, so a strictly-dominated self-pair could be double-ticked into a push) and a real safety hole (the "self-pair" proof rested on the unauthenticated, client-set `written_by` string).

**Round 2 (v2, `fe69adb`): both reviewers returned approve-with-must-fix**, and both confirmed the two round-1 defects are genuinely closed — C-MF1 structurally (`stored_ordering` computed once pre-tick, `else if` exclusivity, ordering passed as a parameter) and C-MF2 by the content-identity clause `tracked.blake3 == remote_hash`, re-derived by hand against the code.

**Round 3 (this head): all five v2 must-fixes applied.** Two commits on top of `fe69adb`, still one file.

### What changed at the gate

1. **"Prove the fix is not a no-op on its own target population"** (the sharpest finding — the re-tick lives only in the regular-file comparator, and `compare_both_exist` routes symlinks away *before* any of the new code runs). Answered with live read-only evidence rather than argument: **all 9 conflicted entries in neo's state cache are regular files — zero symlinks.** `lstat` on the entry keys in `~/.local/share/tcfsd/state.json`; no writes, no daemon interaction, deploy freeze respected. The 7 stuck-at-`times_recorded = 4696` paths are the four `secrets/api/*.age`, `secrets/infrastructure/tailscale_auth_key.age`, `secrets/.manifest.toml`, and `dotfiles/tcfs/devices.json`. Bonus: `secrets/.audit.log` shows up **twice** under two key namespaces with different `blake3` — the live TIN-3278 duplication, confirmed in the wild. Symlink parity remains a correctly-scoped follow-up.

2. **Degenerate-local-content guard** — the residual half of S-MF4, previously answered with prose and a human landing gate. Now code: refuse when `local_size == 0` while `tracked.size > 0`, and for `*.age` paths refuse bytes that no longer carry an age container header (binary **or** armored). Keyed off the `.age` **extension**, not a `secrets/` prefix, because `secrets/.manifest.toml` is plaintext TOML by design and is one of the 7 live stuck paths — a prefix rule would have made this fix a permanent no-op there, which is the very failure mode finding 1 is about. Checked against the live corpus: the 5 stuck `.age` files carry the binary header, so the guard admits them. Residual limit documented, not hidden: header-preserving truncation is not detectable from content alone.

3. **Plan/execute tracked-state divergence** — planning resolved `tracked` with the fuzzy suffix matcher while the push re-resolves its baseline with the exact canonical key, and only that exact entry is ever written back, so under TIN-3278 the safety proof could be evaluated against a record execute never touches. The claim that this "fails closed" was asserted, not proven. Now proven instead: `reconcile()` resolves the exact-key entry too and proves the two lookups return the **same record** (pointer equality into the one entry map); the predicate takes `tracked_is_exact` and refuses when it does not hold, with a `warn!` naming TIN-3278. Nothing else consumes the flag. The full-cycle test — which drives the real `reconcile → execute_plan → state` path — still heals, which is what proves this guard is not itself a no-op.

4. **The new skipped-push log stops asserting a veto.** `upload.skipped` is also set for benign content dedup and `RemoteNewer`, so the message is now keyed off the engine's own `upload.outcome`.

5. **Audit-log honesty.** The `displaced_*` fields are read at plan time while the commit happens after the engine's own fresh remote read, so they were never provably what was displaced. Renamed `displaced_manifest_at_plan` / `displaced_hash_at_plan` and labelled best-effort plan-time provenance in the code, the doc comment, and the S-MF3 row. Emitting from the commit path would mean plumbing through the engine's upload — noted as a follow-up rather than smuggled in.

Also from the non-blocking notes: the doc comment no longer claims the content clause **provably** prevents loss (`validate_indexed_manifest_binding` binds manifest *bytes* to `manifest_hash`, not the declared `file_hash` to the chunk payload), and the S-MF2 disposition is corrected from "closed" to **narrowed, not closed** — the content clause closes the clobber property, not the different claim that the Ed25519-signed device registry stays correctly reflected.

### Evidence the new guards are load-bearing (mutation testing on sting)

| Mutation | Result |
|---|---|
| **C** — delete the `tracked_is_exact` clause | `..._inexact_tracked_lookup_declines_self_heal` **FAILS**: `got Ok(Push { reason: LocalNewer })` |
| **D** — delete both degenerate-content clauses | **two** FAIL: `..._emptied_local_rewrite_is_not_pushed`, `..._non_ciphertext_local_age_rewrite_is_not_pushed` |
| **E** — restore v1's sequential structure with the content clause **intact** | `..._same_content_republish_dominated_clock_is_not_double_ticked` **FAILS**: `got Ok(Push { reason: LocalNewer })` |

Mutation E closes a coverage gap the reviewers named: v2 had no test that pinned the `else if` **independently** of the content clause (mutation B needed both edits). It does now.

### Verification (sting only — never neo)

`cargo test -p tcfs-sync --lib tin3277` → **13 passed, 0 failed**. Full crate: **449 lib tests + 13 integration suites, every one `test result: ok`, 0 failed anywhere**. `cargo fmt --all -- --check` clean, `cargo clippy -p tcfs-sync --all-targets -- -D warnings` clean. Scratch worktree removed.

### Still blocking, by design

The **neo↔honey ciphertext parity check for the 7 stuck paths** stays a landing gate, not a follow-up: age encryption is non-deterministic, so size parity is a weak proxy — decrypt-and-compare or recipient-set comparison is the real check. And the honest limitation is unchanged: this heals the **self**-pair only. The out-of-band writer runs on every host, so expect some conflict volume to migrate to peers rather than drop fleet-wide.

Jesssullivan:
2026-08-15 provenance/CI hold: all four unique commits through `6453eb7abfaa3294f97aa38f800194d99c2298de` are GitHub-unverified/unsigned, and the recorded greens ran on GitHub-hosted platforms. A later signed merge cannot repair that ancestry. Keep this PR unmerged; after #572 and #565 land, recreate the four stable patches as signed commits from frozen post-#565 main, open a replacement PR, and close #576 as superseded. Fresh natural `tinyland-*` CI and the separately attended ciphertext/recipient parity proof are required. No rebase/force/rerun/runtime action.

PR #577, opened 2026-07-30
fix(sync): canonicalize state-cache keys and merge duplicate entries (TIN-3278)

Jesssullivan:
## Current delivery hold — 2026-08-18

This is a dated design/source packet, not a landable exact-head authority. The current PR head is `fee22353df9bf116828c2ec371e380e7652978e9`: its sole commit is unsigned. Its merge base is `e7bfc1f01c3298c5ed6ad25254f6cc67d2a59213`; protected main is now `dfe8282a4d93af0ca1c495065ec82002352d4d86`, so the branch is 1 ahead / 1 behind and its historical checks are not final acceptance evidence.

After #572, #565, and the signed #576 replacement land, reconstruct this patch semantically in a fresh clean worktree on the exact then-current base, create signed non-force replacement history, obtain independent exact-tree review, and require fresh natural sanctioned GloriousFlywheel CI. Do not rerun the old head, apply the migration, touch live state, or treat the Sting test transcript below as current runtime authority. This PR remains Draft with auto-merge off.

## Round-3 design decision: **Option B — reporting collapses, only an explicit locked operation repairs**

Rounds 1 and 2 both implemented this as a duplicate-key **fold at load time**
(`StateCache::open` + `reload_from_disk`). Round 1 found two defects in that
shape; round 2 found four more. Every one of the six is a consequence of the
same decision — mutating cache content on a path that unlocked readers enter —
not of six independent mistakes. Two consecutive rounds of new holes in
implicit-write-at-open is the signal that the design is fighting the
codebase's lock/flush/reload discipline rather than using it.

So round 3 does not add a seventh guard. It removes the load-time fold
entirely and splits the problem along the line the codebase already draws:

| | |
| --- | --- |
| **Reporting** (`StateCache::conflicts_report`) | Collapses duplicate rows per logical file at **read** time. `&self`, borrowed `SyncState`s verbatim, no merge / graft / re-key / delete. Zero new write paths. |
| **Repair** (`StateCache::migrate_duplicate_keys_locked`) | The only thing that mutates. Requires a `StateFileLock` **for this cache**, invoked deliberately via `tcfs state-migrate-keys`. Because it is explicit and locked, it can afford the full recency / refusal guards. |

`open` and `reload_from_disk` are unchanged apart from comments recording why
they must not repair.

### Why the split is the fix and not just a relocation

The user-facing defect is a **reporting** defect: `tcfs conflicts` prints 9
where the daemon's per-cycle plan line reports `conflicts=8`. Nothing about
that symptom requires writing to the cache. Once reporting stops depending on
repair, repair stops needing to happen implicitly, and the two round-2 defect
classes that exist *only because* the fold ran implicitly disappear rather
than being guarded:

- there is no fold on the `recovered_from_backup` flush path, because there is
  no fold on `open` at all (round-2 #2);
- there is no fold on `reload_from_disk`, so the documented "in-memory wins"
  contract is untouched by this PR (round-2 #3).

The remaining two classes shrink to the explicit locked operation, where they
are handled by a **fold invariant** strong enough that they cannot occur:

> The **surviving key** — the key `path_key()` derives, the one a live sync
> cycle has been maintaining — is authoritative for every value-bearing field.
> The fold may only (1) join vector clocks pointwise, (2) raise
> `conflict.times_recorded`, (3) make the duplicate key disappear.

It never writes `blake3` / `size` / `mtime` / `chunk_count` / `remote_path` /
`device_id` / `status`, never writes any conflict payload value (including
`remote_manifest_key` and `rel_path`), and **never gives a key a conflict it
did not already carry**. Note what this replaces: v2 preferred the higher
`last_synced` side, which is wall-clock and skew-prone, and backfilled
`remote_manifest_key` across the fold. Both are gone.

Two refusals back it up, reported to the operator rather than guessed at:

- `WouldResurrectConflict` — the survivor is clean and the duplicate carries a
  conflict. **Refuse the fold entirely**; both keys stay. No record is lost and
  no clean entry becomes conflicted (round-2 #1). Stronger than the requested
  recency guard, which would still have grafted when `detected_at` was newer.
- `DuplicateConflictIsNewer` — the duplicate's conflict post-dates the
  survivor's, contradicting the "duplicate is frozen history" premise. Refuse;
  an operator should look.

### What Option A would and would not have bought

Honest accounting, since Option A (patch the fold against the round-2
six-point checklist) was the alternative:

| Round-2 class | Option A | Option B (this PR) |
| --- | --- | --- |
| 1 — conflict resurrection | guarded (recency check on the graft) | **eliminated** (fold refuses outright; nothing is ever grafted) |
| 2 — `.bak`-recovery unlocked write | guarded (skip fold when `recovered_from_backup`) | **eliminated** (no fold on `open`) |
| 3 — `reload_from_disk` contract break | guarded (exclude reload, or non-overwriting fold) | **eliminated** (no fold on reload) |
| 4 — git-consumer widening | argued + tested per consumer | **eliminated** (invariant: no key can gain those values or a conflict) |

Option A leaves four live guards on an implicit write path, each of which must
stay correct as the fold and its callers evolve, and each of which is a place a
round 4 can find a hole. Option B leaves one write path that a caller must ask
for by name and hold a lock to reach. Only Option B has a structural argument;
Option A has four behavioral ones.

Cost of B, stated plainly: the repair is no longer automatic. An operator has
to run `tcfs state-migrate-keys --execute`. That is the trade, and it is the
right one for a cache whose corruption mode is "an entry no live path touches."

---

## Round-2 findings → disposition

| # | Round-2 must-fix | Disposition |
| --- | --- | --- |
| 1 | Conflict resurrection: stale conflict grafted onto a live clean entry, no recency guard; the PR's own test locked the defect in | **ELIMINATED.** The fold refuses when only the duplicate carries a conflict (`WouldResurrectConflict`) and when the duplicate's conflict is newer (`DuplicateConflictIsNewer`). No graft path exists at any `detected_at`. The offending test is gone; `tin3278_explicit_migration_refuses_to_resurrect_a_stale_conflict` asserts the refusal, that both keys survive, and that the clean entry stays `Synced` |
| 2 | `recovered_from_backup` still makes unlocked readers flush the **migrated** fold; `dirty:false` does not help; the regression test never writes a `.bak` | **ELIMINATED.** No fold on `open`, so the recovery flush writes back what it read. `tin3278_bak_recovery_open_writes_back_unfolded_content` builds the real shape (corrupt primary + valid `.bak` + duplicate pair), asserts `recovered_from_backup`, drops the cache so the recovery flush fires unlocked, then re-loads and asserts the duplicate key survived and no entry merged. The vacuous `debug_assert` is deleted |
| 3 | `reload_from_disk` breaks its documented "in-memory wins" contract; orphan re-inserted every reload, fold overwrites live values on a wall-clock tiebreak | **ELIMINATED.** No fold on reload; `or_insert` is again the whole contract. `tin3278_reload_from_disk_keeps_in_memory_values` runs the exact scenario from the finding — in-memory `blake3=NEW`/`last_synced=1000` vs disk `blake3=OLD`/`last_synced=5000` — and asserts the live entry keeps `NEW` |
| 4 | Grafted legacy-prefix `remote_manifest_key` makes `collect_repo_conflicts` bail the whole repo resolution; also flips daemon git routing at `grpc.rs:4296` | **ELIMINATED.** No backfill, no `status` forcing. `tin3278_explicit_migration_cannot_widen_git_consumer_fields` builds a `.git`-internal pair whose stale duplicate carries a **higher** `last_synced` and legacy prefixes in *both* `remote_path` and `remote_manifest_key`, plus a clean/conflicted refusal pair, and asserts: the set of conflicted keys after the fold is a subset of the set before; the survivor's `status`, `remote_path`, `remote_manifest_key`, `rel_path`, `blake3` are byte-identical; `times_recorded` rose to 1881; no folded key carries a `legacy/` address; and the refused duplicate is untouched |
| 5 (non-blocking) | `conflicts_naming_unknown_devices` has zero callers; "defect 2 fixed" framing unearned | **DROPPED** from the PR — function and tests removed. The ghost-device reporting scan is not claimed here |
| — | Pass-2 cost: O(orphans × entries) with per-candidate `normalize_rel_path_text` on the `open`/reload path; one `canonicalize` syscall per key per CLI invocation | **ELIMINATED from the hot path.** `open` and `reload_from_disk` issue **zero** extra syscalls — the migration does not run there. The fold's cost is paid only when an operator invokes the verb. `conflicts_report` canonicalizes once per **conflict** row (the small minority of a cache), not per entry |
| — | Dead `normalized_candidate == normalized_orphan` branch | **REMOVED** |
| — | `tin3278_migration_is_idempotent_on_second_load` only passed via an artificial `set`+`flush` | **REPOINTED.** `tin3278_explicit_migration_is_idempotent` folds under lock, immediately re-runs the same locked migration and asserts the plan is empty, then reopens the persisted file and asserts `plan_duplicate_key_migration()` is empty. No injected write |
| — | Case-sensitivity on APFS unproven | **DOCUMENTED** as a declining case on `fold_duplicate_keys`; matching is byte-exact and an orphan differing only by case is left in place and keeps double-counting. Not silently claimed as fixed |
| — | Multi-root boundary wants its own ticket | **REPORTED, not silent.** 2+ resolvable candidates now surface as `KeyMigrationSkipReason::NoUnambiguousTarget` in the plan the operator reads before `--execute`; `tin3278_explicit_migration_reports_ambiguous_multi_root_orphan` covers it. Still a boundary — state-cache keys carry no root attribution |

---

## What changed, concretely

**`crates/tcfs-sync/src/state.rs`**

- `StateFileLock` carries the state-cache path it guards, exposed via
  `state_path()`. So an API that *requires* a held lock can verify the caller
  handed it a lock for the right cache instead of trusting that any guard in
  scope is the relevant one.
- `StateCache::conflicts_report() -> Vec<ConflictReportRow<'_>>` — read-only
  collapse. Rules, deliberately conservative:
  - rows whose keys resolve on disk to the same path are one group; the
    representative is the row already stored under the resolved path when
    present, else highest `last_synced`, ties broken on key order (deterministic
    output — `conflicts()` iterates a `HashMap`);
  - an unresolvable orphan attaches to a resolvable row by the same rel-path
    suffix convention `get_by_rel_path` already uses, and **only** on exactly
    one match; 0 or 2+ leave it as its own visible row;
  - **a conflict with no duplicate is never suppressed** — an orphan conflict
    whose live counterpart is clean stays visible. It is real cache debt, and
    hiding it is worse than double-counting.
- `plan_duplicate_key_migration()` — pure dry run; clones the map and runs the
  *identical* routine as the apply step, so plan and apply cannot disagree.
- `migrate_duplicate_keys_locked(&StateFileLock)` — the sole repair path;
  verifies the lock is for this cache, folds, dirties, flushes under the
  caller's still-live guard, returns the audit trail.
- `merge_duplicate_sync_states` returns a three-way `DuplicateFold`
  (`Folded` / `RefusedWouldResurrectConflict` / `RefusedDuplicateConflictIsNewer`)
  instead of always producing a merged state.
- Vector clocks are still **joined, not picked** (round-1 #1, unchanged and
  still correct): pointwise `VectorClock::merge` on the entry clock, and
  strictly same-side on the payload (`local`∨`local`, `remote`∨`remote`, never
  crossed). `partial_cmp_vc` reads an absent component as `0`, so dropping one
  makes a genuinely concurrent pair read as *remote* dominance → false
  `RemoteNewer` → silent overwrite. Two tests pin the direction.
- `conflicts_naming_unknown_devices` and its tests removed.

**`crates/tcfs-cli/src/main.rs`**

- `tcfs conflicts` (primary/legacy offline branch) reports from
  `conflicts_report()`. Same command, no lock taken, still writes nothing —
  it now prints the per-logical-file count that matches the daemon, names each
  shadowed duplicate key, and points at the repair verb. `--json` gains
  `duplicate_cache_keys`.
- `tcfs state-migrate-keys [--execute] [--json] [--state <path>]` — dry run by
  default. **Both** modes take the `StateFileLock` before `StateCache::open`,
  not just `--execute`: `open` itself can write, because it recovers a missing
  or corrupt primary from `.bak` and `Drop` flushes that recovery even for a
  nominally read-only command — the hazard `lock_explicit_state_cache`
  documents and round-2 #2 caught. Taking the lock up front means this verb can
  never write unlocked in either mode, and a run colliding with a daemon cycle
  fails loudly instead of racing it.

### Precision about pass-1 re-keying (stated so it isn't found later)

`merge_duplicate_sync_states` cannot give an *existing* key a conflict it did
not carry. A pass-1 **pure** re-key (destination key absent) does move an
existing conflict onto a key the cache did not previously hold, which
`grpc.rs`'s `Path::new(cache_key).starts_with(git_dir)` could then match. That
is not widening: the destination is by construction the key `path_key()`
derives for the very same file (same `file_name`, canonicalized parent), so the
entry lands exactly where a live `get`/`set`/`mark_conflict` was always going to
look. Routing follows the live reconciler instead of diverging from it — the
mismatch this repairs is the whole bug. Recorded on `fold_duplicate_keys`.

### Known boundary: the daemon's registered-root conflict list

Unchanged and out of scope. `tcfs conflicts --root <id>` is served by the
daemon's `conflict_records` (`tcfsd/src/grpc.rs`), which scans `conflicts()`
raw. That path is already prefix-validated per root by
`validate_conflict_cache_route`, and the observed 9-vs-8 divergence is on the
primary/legacy cache the offline branch reads — so wiring `conflicts_report()`
into the daemon is a separate change in a separate crate, deliberately not
bundled here.

---

## Tests — 17 TIN-3278 tests

Load-path guarantees (the ones round 2 asked for):
`tin3278_open_never_folds_or_rewrites_duplicate_keys` ·
`tin3278_bak_recovery_open_writes_back_unfolded_content` ·
`tin3278_reload_from_disk_keeps_in_memory_values`

Read-side collapse:
`tin3278_conflicts_report_collapses_duplicate_rows_without_mutating` ·
`tin3278_conflicts_report_never_hides_a_lone_orphan_conflict` ·
`tin3278_conflicts_report_leaves_ambiguous_orphan_visible` ·
`tin3278_conflicts_report_collapses_resolvable_alias_key`

Explicit locked migration:
`tin3278_explicit_migration_requires_a_lock_for_this_cache` ·
`tin3278_explicit_migration_dry_run_plan_does_not_mutate` ·
`tin3278_explicit_migration_folds_and_persists_under_lock` ·
`tin3278_explicit_migration_refuses_to_resurrect_a_stale_conflict` ·
`tin3278_explicit_migration_refuses_when_duplicate_conflict_is_newer` ·
`tin3278_explicit_migration_reports_ambiguous_multi_root_orphan` ·
`tin3278_explicit_migration_cannot_widen_git_consumer_fields` ·
`tin3278_explicit_migration_is_idempotent`

Fold semantics:
`tin3278_fold_joins_vector_clocks_instead_of_dropping_one_side` ·
`tin3278_fold_joins_conflict_side_clocks_same_side_only`

## Gates — sting, `/tmp/tin3278-wt4` @ `fee22353`, cargo 1.91.0 (worktree removed after)

| Gate | Result |
| --- | --- |
| `cargo fmt --all --check` | **PASS** (exit 0) — no diffs |
| `cargo test -p tcfs-sync` | **PASS** (exit 0) — 579 passed, 0 failed, 0 ignored |
| `cargo clippy -p tcfs-sync --all-targets -- -D warnings` | **PASS** (exit 0) — 0 warnings |
| `cargo clippy -p tcfs-cli --all-targets` | **PASS** — 0 diagnostics attributed to `crates/tcfs-cli` (see caveat) |

Caveat on the last row, stated because it is a weaker gate than the other
three: `-p tcfs-cli` **cannot** be resolved on sting at all — `async-notify
v0.3.4` requires rustc 1.92.0 and sting is on 1.91.1. Verified pre-existing:
the identical failure reproduces on a clean `origin/main` worktree, so it is
not introduced here. To typecheck the CLI change anyway it was built with a
**worktree-local** `cargo update async-notify --precise 0.3.2`; `Cargo.lock` is
**not** modified in this PR and the three primary gates above all ran against
the pristine lockfile *before* that pin. The only diagnostic in that build is a
pre-existing `clippy::only_used_in_recursion` warning in
`crates/tcfs-secrets/src/kdbx.rs`, untouched by this PR.

## Test plan

- [x] `cargo fmt --all --check` on sting
- [x] `cargo test -p tcfs-sync` on sting
- [x] `cargo clippy -p tcfs-sync --all-targets -- -D warnings` on sting
- [x] `cargo clippy -p tcfs-cli --all-targets` on sting (local dep pin; see caveat)
- [ ] Manual, on a **read-only copy** of neo's `state.json` (LAB_DEPLOY_FREEZE:
      copy only, no live state mutation): confirm `tcfs conflicts` collapses the
      `/secrets/.audit.log` duplicate and matches the daemon's cycle count, that
      the copy is byte-identical afterwards, and that `tcfs state-migrate-keys`
      dry-run plans exactly that one fold

## Coordination

- **#576 / TIN-3277** landed a `tracked_is_exact` guard in `reconcile.rs` that
  warns when a lookup resolves through a dup-key path. Still complementary:
  that guard reports, `tcfs state-migrate-keys` removes. Disjoint files
  (`reconcile.rs` vs `state.rs` + `tcfs-cli/main.rs`).
- **#565 / TIN-2864** (Codex-owned, not modified) touches an unrelated region of
  `state.rs` (bounded-read primitives near the top); no construct-level overlap.

Stays **DRAFT** pending the round-3 gate.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -392,6 +392,29 @@ enum Commands {
         state: Option<PathBuf>,
     },
 
+    /// Repair duplicate state-cache key namespaces (TIN-3278)
+    ///
+    /// One logical file can be tracked under two keys — the absolute
+    /// canonicalized path live reads/writes derive, plus a bare
+    /// prefix-relative key left by an older keying scheme — which
+    /// double-counts in raw scans such as `tcfs conflicts`. This is the only
+    /// verb that repairs that: it takes the state-cache lock and folds the
+    /// duplicate away. Dry run unless `--execute` is given.
+    #[command(name = "state-migrate-keys")]
+    StateMigrateKeys {
+        /// Apply the migration under the state-cache lock. Without this flag
+        /// nothing is written.
+        #[arg(long)]
+        execute: bool,
+        /// Emit machine-readable JSON instead of the human summary
+        #[arg(long)]
+        json: bool,
+        /// Path to the sync state cache JSON file (overrides config).
+        /// `.db` paths are normalized to their `.json` sibling.
+        #[arg(long, env = "TCFS_STATE_PATH")]
+        state: Option<PathBuf>,
+    },
+
     /// Manage the sync trash (staged deletes)
     ///
     /// When trash is enabled, deletion writes an immutable safety copy and
@@ -1079,6 +1102,11 @@ async fn main() -> Result<()> {
         Commands::Conflicts { json, root, state } => {
             cmd_conflicts(&config, json, root.as_deref(), state.as_deref()).await
         }
+        Commands::StateMigrateKeys {
+            execute,
+            json,
+            state,
+        } => cmd_state_migrate_keys(&config, execute, json, state.as_deref()).await,
     }
 }
 
@@ -9151,13 +9179,142 @@ fn conflict_age(detected_at: u64) -> String {
 /// `tcfs conflicts` — list recorded conflicts, grouped by repo for
 /// `.git`-internal paths. Named roots use the daemon-selected cache; the
 /// primary/legacy path stays an offline read.
+/// Tell the operator when the conflict list collapsed duplicate key-namespace
+/// rows, and how to repair the cache (TIN-3278).
+///
+/// Reporting deliberately does not repair: `tcfs conflicts` on the default
+/// primary holds no `StateFileLock`, so it must not write.
+fn report_duplicate_cache_keys(duplicate_cache_keys: &[String], state_path: &Path) {
+    if duplicate_cache_keys.is_empty() {
+        return;
+    }
+    println!(
+        "note: {} duplicate cache key(s) collapsed for reporting (TIN-3278); \
+         the count above is per logical file, matching the daemon's cycle count.",
+        duplicate_cache_keys.len()
+    );
+    for key in duplicate_cache_keys {
+        println!("  duplicate key: {key}");
+    }
+    println!(
+        "  repair (writes, takes the state lock): tcfs state-migrate-keys --state {} --execute",
+        state_path.display()
+    );
+}
+
+/// `tcfs state-migrate-keys` — explicit duplicate-key repair (TIN-3278).
+///
+/// Dry run by default. **Both** modes acquire the same `StateFileLock` every
+/// other state-cache writer uses, before `StateCache::open`. The dry run needs
+/// it too, and not for the migration (which it never calls): `open` itself can
+/// write, because it recovers a missing or corrupt primary from `.bak` and
+/// `Drop` flushes that recovery even for a nominally read-only command — the
+/// same hazard `lock_explicit_state_cache` documents. Taking the lock up front
+/// means this verb can never write unlocked in either mode, and a run that
+/// collides with a daemon cycle fails loudly instead of racing it.
+///
+/// `--execute` hands the guard to `StateCache::migrate_duplicate_keys_locked`,
+/// which verifies it is a lock for *this* cache and refuses otherwise. This is
+/// the only path that repairs duplicate key namespaces; `open` and
+/// `reload_from_disk` never do.
+async fn cmd_state_migrate_keys(
+    config: &tcfs_core::config::TcfsConfig,
+    execute: bool,
+    json: bool,
+    state_override: Option<&Path>,
+) -> Result<()> {
+    let state_path = resolve_state_path(config, state_override);
+
+    let render = |plan: &tcfs_sync::state::KeyMigrationPlan| {
+        if json {
+            let merges: Vec<serde_json::Value> = plan
+                .merges
+                .iter()
+                .map(|record| {
+                    serde_json::json!({
+                        "dropped_key": record.dropped_key,
+                        "kept_key": record.kept_key,
+                        "conflict_preserved": record.conflict_preserved,
+                    })
+                })
+                .collect();
+            let skipped: Vec<serde_json::Value> = plan
+                .skipped
+                .iter()
+                .map(|(key, reason)| {
+                    serde_json::json!({ "key": key, "reason": format!("{reason:?}") })
+                })
+                .collect();
+            println!(
+                "{}",
+                serde_json::to_string_pretty(&serde_json::json!({
+                    "state_path": state_path.to_string_lossy(),
+                    "executed": execute,
+                    "merges": merges,
+                    "skipped": skipped,
+                }))
+                .unwrap_or_else(|_| "{}".to_string())
+            );
+            return;
+        }
+        println!("state cache: {}", state_path.display());
+        println!(
+            "{} ({} fold(s), {} left in place)",
+            if execute {
+                "applied"
+            } else {
+                "dry run — nothing written"
+            },
+            plan.merges.len(),
+            plan.skipped.len()
+        );
+        for record in &plan.merges {
+            println!(
+                "  fold {} -> {}{}",
+                record.dropped_key,
+                record.kept_key,
+                if record.conflict_preserved {
+                    " (conflict recurrence/clocks absorbed)"
+                } else {
+                    ""
+                }
+            );
+        }
+        for (key, reason) in &plan.skipped {
+            println!("  keep {key}: {reason:?}");
+        }
+        if !execute && !plan.merges.is_empty() {
+            println!("re-run with --execute to apply under the state-cache lock.");
+        }
+    };
+
+    // Held for both modes, and taken before `open` — see the note above.
+    let lock = tcfs_sync::state::StateFileLock::acquire(&state_path).with_context(|| {
+        format!(
+            "locking state cache for migration: {}",
+            state_path.display()
+        )
+    })?;
+    let mut cache = tcfs_sync::state::StateCache::open(&state_path)
+        .with_context(|| format!("opening state cache: {}", state_path.display()))?;
+
+    if !execute {
+        render(&cache.plan_duplicate_key_migration());
+        return Ok(());
+    }
+
+    let applied = cache.migrate_duplicate_keys_locked(&lock)?;
+    render(&applied);
+    Ok(())
+}
+
 async fn cmd_conflicts(
     config: &tcfs_core::config::TcfsConfig,
     json: bool,
     root_id: Option<&str>,
     state_override: Option<&Path>,
 ) -> Result<()> {
-    let (state_path, routed_local_root, routed_remote_prefix, items) =
+    let (state_path, routed_local_root, routed_remote_prefix, items, duplicate_cache_keys) =
         if let Some(root_id) = root_id {
             let mut client = connect_daemon(&config.daemon.socket).await?;
             let response = client
@@ -9201,23 +9358,35 @@ async fn cmd_conflicts(
                 Some(response.local_root),
                 Some(response.remote_prefix),
                 items,
+                Vec::new(),
             )
         } else {
             let state_path = resolve_state_path(config, state_override);
             let _state_lock = lock_explicit_state_cache(&state_path, state_override)?;
             let state = tcfs_sync::state::StateCache::open(&state_path)
                 .with_context(|| format!("opening state cache: {}", state_path.display()))?;
-            let items: Vec<(String, tcfs_sync::conflict::ConflictInfo)> = state
-                .conflicts()
-                .into_iter()
-                .filter_map(|(key, state)| {
-                    state
+            // TIN-3278: collapse duplicate key-namespace rows for the same
+            // logical file at *read* time. This is reporting only — it never
+            // mutates the cache, so a `tcfs conflicts` that holds no state-file
+            // lock (the default primary path: `lock_explicit_state_cache` takes
+            // the lock only for an explicit `--state`) still writes nothing.
+            // Repair is the separate, locked `tcfs state-migrate-keys` verb.
+            let report = state.conflicts_report();
+            let mut duplicate_keys: Vec<String> = report
+                .iter()
+                .flat_map(|row| row.shadowed_keys.iter().map(|key| (*key).to_string()))
+                .collect();
+            duplicate_keys.sort();
+            let items: Vec<(String, tcfs_sync::conflict::ConflictInfo)> = report
+                .iter()
+                .filter_map(|row| {
+                    row.state
                         .conflict
                         .as_ref()
-                        .map(|conflict| (key.to_string(), conflict.clone()))
+                        .map(|conflict| (row.cache_key.to_string(), conflict.clone()))
                 })
                 .collect();
-            (state_path, None, None, items)
+            (state_path, None, None, items, duplicate_keys)
         };
 
     let groups = group_conflicts(&items);
@@ -9257,6 +9426,11 @@ async fn cmd_conflicts(
                 "remote_prefix": routed_remote_prefix.as_deref(),
                 "state_path": state_path.to_string_lossy(),
                 "conflict_count": items.len(),
+                // TIN-3278: duplicate key-namespace rows collapsed for
+                // reporting. Present means the cache carries repairable debt;
+                // the count above is the per-logical-file count the reconciler
+                // acts on.
+                "duplicate_cache_keys": duplicate_cache_keys,
                 "groups": rendered,
             }))?
         );
@@ -9268,6 +9442,7 @@ async fn cmd_conflicts(
             println!("Root: {root_id}");
         }
         println!("No recorded conflicts.");
+        report_duplicate_cache_keys(&duplicate_cache_keys, &state_path);
         return Ok(());
     }
 
@@ -9286,6 +9461,7 @@ async fn cmd_conflicts(
         items.len(),
         groups.len()
     );
+    report_duplicate_cache_keys(&duplicate_cache_keys, &state_path);
     for g in &groups {
         println!();
         match &g.repo_root {
--- a/crates/tcfs-sync/src/state.rs
+++ b/crates/tcfs-sync/src/state.rs
@@ -277,6 +277,13 @@ fn sync_parent_directory(_path: &Path) -> Result<()> {
 #[derive(Debug)]
 pub struct StateFileLock {
     _file: std::fs::File,
+    /// The state cache this lock guards (not the `.lock` sibling).
+    ///
+    /// Carried so an API that *requires* a held lock can verify the caller
+    /// handed it a lock for the right cache instead of trusting that any lock
+    /// guard in scope is the relevant one — see
+    /// [`StateCache::migrate_duplicate_keys_locked`].
+    state_path: PathBuf,
 }
 
 /// Result of probing the existing writer lock without creating or modifying it.
@@ -306,6 +313,11 @@ impl Drop for StateFileLock {
 }
 
 impl StateFileLock {
+    /// The state cache this guard locks.
+    pub fn state_path(&self) -> &Path {
+        &self.state_path
+    }
+
     /// Stable sibling path used by every state-cache writer.
     pub fn lock_path(state_path: &Path) -> PathBuf {
         let mut path = state_path.as_os_str().to_os_string();
@@ -350,7 +362,10 @@ impl StateFileLock {
         validate_opened_state_file(&file, &lock_path)
             .with_context(|| format!("revalidating opened state lock: {}", lock_path.display()))?;
         match file.try_lock() {
-            Ok(()) => Ok(Self { _file: file }),
+            Ok(()) => Ok(Self {
+                _file: file,
+                state_path: state_path.to_path_buf(),
+            }),
             Err(std::fs::TryLockError::WouldBlock) => anyhow::bail!(
                 "state cache {} is locked by another process; retry after the current reconcile cycle (lock: {})",
                 state_path.display(),
@@ -399,7 +414,10 @@ impl StateFileLock {
             .with_context(|| format!("validating existing state lock: {}", lock_path.display()))?;
 
         match file.try_lock() {
-            Ok(()) => Ok(ExistingStateFileLock::Acquired(Self { _file: file })),
+            Ok(()) => Ok(ExistingStateFileLock::Acquired(Self {
+                _file: file,
+                state_path: state_path.to_path_buf(),
+            })),
             Err(std::fs::TryLockError::WouldBlock) => Ok(ExistingStateFileLock::Contended),
             Err(std::fs::TryLockError::Error(error)) => Err(error).with_context(|| {
                 format!("locking existing state cache via {}", lock_path.display())
@@ -697,6 +715,94 @@ pub struct StateCacheKeySnapshot {
     recovered_from_backup: bool,
 }
 
+/// One conflict row as it should be *reported* to an operator (TIN-3278).
+///
+/// Produced by [`StateCache::conflicts_report`]. `state` is borrowed straight
+/// out of the cache — nothing about it is merged or synthesized.
+#[derive(Debug, Clone)]
+pub struct ConflictReportRow<'a> {
+    /// The key that represents this logical file: the one live
+    /// `get`/`set`/`mark_conflict` derive whenever it is present in the cache.
+    pub cache_key: &'a str,
+    /// The entry recorded under [`ConflictReportRow::cache_key`], verbatim.
+    pub state: &'a SyncState,
+    /// Duplicate keys for the same logical file that this row stands in for.
+    /// Non-empty means the cache carries key-namespace debt an operator can
+    /// repair with the explicit migration; the rows themselves were suppressed
+    /// only for reporting.
+    pub shadowed_keys: Vec<&'a str>,
+}
+
+/// One duplicate-key fold performed by an explicit, lock-holding duplicate-key
+/// migration (TIN-3278).
+///
+/// The primary state cache historically accumulated entries for the same
+/// logical file under two key namespaces — an absolute canonicalized path
+/// (the form [`path_key`] always produces today) and a bare prefix-relative
+/// path left behind by an older keying scheme. The orphaned relative key is
+/// never touched by any live read/write path (`get`/`set`/`mark_conflict`
+/// all re-derive the canonical key via [`path_key`]), so once such an entry
+/// exists it lingers forever, double-counting in raw-entry scans like
+/// [`StateCache::conflicts`] (`tcfs conflicts`) even though the live
+/// reconciler only ever visits the canonical path.
+///
+/// A record is produced per fold, both by the dry-run planner
+/// ([`StateCache::plan_duplicate_key_migration`]) and by the apply step
+/// ([`StateCache::migrate_duplicate_keys_locked`]); it is purely an audit
+/// trail (`Debug`/`Clone`, no `Drop` side effects).
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct KeyMergeRecord {
+    /// The non-canonical (or otherwise duplicate) key whose entry was folded
+    /// away. Removed from the cache once the merge completes.
+    pub dropped_key: String,
+    /// The canonical key that survived the merge and now carries the
+    /// combined entry.
+    pub kept_key: String,
+    /// Set when the dropped side carried an active `conflict` record whose
+    /// monotone parts (`times_recorded`, vector clocks) were folded into the
+    /// surviving key's own conflict payload. The surviving key's payload
+    /// **values** are never replaced — see [`merge_duplicate_sync_states`].
+    pub conflict_preserved: bool,
+}
+
+/// Why an explicit duplicate-key migration declined to fold a key (TIN-3278).
+///
+/// Every skip leaves the cache exactly as it was, so a skipped key keeps
+/// double-counting in raw scans; [`StateCache::conflicts_report`] still
+/// collapses it for reporting. Skips are surfaced to the operator rather than
+/// silently guessed at.
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum KeyMigrationSkipReason {
+    /// The orphan matched zero or 2+ independently-resolvable keys by rel-path
+    /// suffix, so there is no unambiguous target (the multi-root case).
+    NoUnambiguousTarget { resolvable_candidates: usize },
+    /// The surviving key carries no conflict and the duplicate does. Folding
+    /// would turn a cosmetic double-count into a live conflict on a currently
+    /// clean entry (and hand that entry's key to `conflict_git` /
+    /// git-routing consumers), so the duplicate is left in place instead.
+    WouldResurrectConflict { kept_key: String },
+    /// The duplicate's conflict payload is *newer* than the surviving key's,
+    /// contradicting the premise that the duplicate is frozen history. Rather
+    /// than choose between two live-looking records, leave both for an operator.
+    DuplicateConflictIsNewer { kept_key: String },
+}
+
+/// What an explicit duplicate-key migration did, or would do (TIN-3278).
+#[derive(Debug, Clone, Default, PartialEq, Eq)]
+pub struct KeyMigrationPlan {
+    /// Folds performed (or, for a dry run, that would be performed).
+    pub merges: Vec<KeyMergeRecord>,
+    /// Duplicates deliberately left alone, with the reason.
+    pub skipped: Vec<(String, KeyMigrationSkipReason)>,
+}
+
+impl KeyMigrationPlan {
+    /// No duplicate key namespaces to repair and nothing declined.
+    pub fn is_empty(&self) -> bool {
+        self.merges.is_empty() && self.skipped.is_empty()
+    }
+}
+
 impl StateCache {
     /// Load or create a state cache at the given path.
     ///
@@ -768,6 +874,27 @@ impl StateCache {
             (HashMap::new(), 0, String::new(), false)
         };
 
+        // TIN-3278: `open` deliberately does NOT repair duplicate key
+        // namespaces. Duplicate-key repair mutates entry content (it deletes a
+        // key and merges two states), and this constructor cannot write:
+        //
+        // - `Drop` flushes whenever `dirty || recovered_from_backup`
+        //   (see `impl Drop for StateCache`), and the `recovered_from_backup`
+        //   leg fires even for a nominally read-only command — which is exactly
+        //   why `lock_explicit_state_cache` (tcfs-cli) exists and says so;
+        // - the CLI takes a cross-process `StateFileLock` only when a `--state`
+        //   override is supplied, so `tcfs conflicts` on the default primary
+        //   holds no lock at all;
+        // - the daemon's startup lock is `<data_dir>/daemon-instance.lock`
+        //   (tcfsd/src/daemon.rs), not this cache's `state.json.lock`.
+        //
+        // A content-mutating fold here would therefore be persisted by unlocked
+        // readers racing the daemon's atomic-rename flush. Repair is an
+        // explicit, lock-holding operation instead:
+        // [`StateCache::plan_duplicate_key_migration`] (pure) and
+        // [`StateCache::migrate_duplicate_keys_locked`] (requires the lock).
+        // Reporting surfaces collapse duplicates at read time with
+        // [`StateCache::conflicts_report`], which never mutates anything.
         Ok(StateCache {
             db_path: db_path.to_path_buf(),
             entries,
@@ -839,6 +966,12 @@ impl StateCache {
         if self.device_id.is_empty() && !device_id.is_empty() {
             self.device_id = device_id;
         }
+        // TIN-3278: no duplicate-key fold here either. The `or_insert` above is
+        // the whole "in-memory wins" contract; a fold would violate it, because
+        // merging a disk entry into a resident key can only be done by choosing
+        // between a disk value and an in-memory value. Duplicate repair is the
+        // explicit locked operation — see `open`'s note and
+        // [`StateCache::migrate_duplicate_keys_locked`].
         Ok(())
     }
 
@@ -936,6 +1069,232 @@ impl StateCache {
             .filter(|(_, s)| s.conflict.is_some())
             .map(|(k, v)| (k.as_str(), v))
             .collect()
+    }
+
+    /// Conflicts collapsed to one row per logical file, for **reporting only**
+    /// (TIN-3278).
+    ///
+    /// [`StateCache::conflicts`] is a raw entry scan, so a file tracked under
+    /// two key namespaces (an absolute canonicalized path — the form
+    /// [`path_key`] always produces on live `get`/`set`/`mark_conflict` — plus a
+    /// bare prefix-relative key left by an older keying scheme) is reported
+    /// twice, while the reconciler only ever visits the canonical key. That is
+    /// the whole user-visible TIN-3278 symptom: `tcfs conflicts` printing 9
+    /// where the daemon's per-cycle plan line reports `conflicts=8`.
+    ///
+    /// This collapses those rows **without touching the cache**: it takes
+    /// `&self`, returns borrowed [`SyncState`]s exactly as recorded, and does
+    /// not merge, graft, re-key, or delete anything. Duplicate repair is a
+    /// separate, explicitly invoked, lock-holding operation
+    /// ([`StateCache::migrate_duplicate_keys_locked`]).
+    ///
+    /// Collapse rules — deliberately conservative:
+    ///
+    /// - Rows whose keys resolve on disk to the *same* path (e.g. `/var/f` and
+    ///   `/private/var/f`, or a symlinked-parent duplicate) are one group. The
+    ///   representative is the row already stored under that resolved path when
+    ///   present; otherwise the highest `last_synced`, breaking ties on the key
+    ///   so output is deterministic.
+    /// - A row whose key cannot resolve at all (the live TIN-3278 orphan shape:
+    ///   `/secrets/.audit.log`, whose `/secrets` does not exist at the
+    ///   filesystem root) is attached to a resolvable row using the same
+    ///   rel-path-suffix convention [`StateCache::get_by_rel_path`] uses for
+    ///   cross-host lookups — but only when **exactly one** resolvable row
+    ///   matches. Zero or 2+ candidates leave the orphan as its own reported
+    ///   row rather than guess.
+    /// - A conflict that has no duplicate among the reported rows is **never**
+    ///   suppressed. In particular, an orphan conflict whose live counterpart is
+    ///   currently clean stays visible: it is real cache debt, and hiding it (or
+    ///   grafting it onto the clean entry) would be worse than double-counting.
+    ///
+    /// Cost: one `fs::canonicalize` attempt per *conflict* row (not per cache
+    /// entry), plus an O(orphan rows x groups) suffix scan over pre-normalized
+    /// strings. Conflict rows are the small minority of a cache.
+    pub fn conflicts_report(&self) -> Vec<ConflictReportRow<'_>> {
+        /// One logical-file group under construction.
+        struct Group<'a> {
+            representative: &'a str,
+            state: &'a SyncState,
+            shadowed: Vec<&'a str>,
+            /// Independently resolved on-disk form, or the literal key when the
+            /// key does not resolve at all.
+            identity: String,
+            normalized_identity: String,
+            resolvable: bool,
+        }
+
+        // Deterministic input order: `conflicts()` iterates a HashMap.
+        let mut rows: Vec<(&str, &SyncState)> = self.conflicts();
+        rows.sort_by(|left, right| left.0.cmp(right.0));
+
+        // Phase 1: group rows that resolve to the same on-disk path.
+        let mut groups: Vec<Group<'_>> = Vec::new();
+        for (cache_key, state) in rows {
+            let resolved = resolve_key_on_disk(cache_key);
+            let identity = resolved.clone().unwrap_or_else(|| cache_key.to_string());
+            match groups.iter().position(|group| group.identity == identity) {
+                Some(index) => {
+                    let group = &mut groups[index];
+                    // Prefer the row already stored under the resolved path —
+                    // that is the key live `get`/`set` derive. Otherwise the
+                    // most recently synced row, ties broken on key order so the
+                    // report is stable.
+                    let challenger_wins =
+                        match (group.representative == identity, cache_key == identity) {
+                            (true, false) => false,
+                            (false, true) => true,
+                            _ => {
+                                state.last_synced > group.state.last_synced
+                                    || (state.last_synced == group.state.last_synced
+                                        && cache_key < group.representative)
+                            }
+                        };
+                    if challenger_wins {
+                        group.shadowed.push(group.representative);
+                        group.representative = cache_key;
+                        group.state = state;
+                    } else {
+                        group.shadowed.push(cache_key);
+                    }
+                }
+                None => groups.push(Group {
+                    representative: cache_key,
+                    state,
+                    shadowed: Vec::new(),
+                    normalized_identity: crate::engine::normalize_rel_path_text(&identity),
+                    resolvable: resolved.is_some(),
+                    identity,
+                }),
+            }
+        }
+
+        // Phase 2: attach an unresolvable orphan to its resolvable counterpart,
+        // but only when exactly one candidate matches by rel-path suffix.
+        let mut fold_target: Vec<Option<usize>> = vec![None; groups.len()];
+        for index in 0..groups.len() {
+            if groups[index].resolvable {
+                continue;
+            }
+            let normalized_orphan = crate::engine::normalize_rel_path_text(
+                groups[index].representative.trim_start_matches('/'),
+            );
+            if normalized_orphan.is_empty() {
+                continue;
+            }
+            let suffix = format!("/{normalized_orphan}");
+            let mut candidates = (0..groups.len()).filter(|other| {
+                *other != index
+                    && groups[*other].resolvable
+                    && groups[*other].normalized_identity.ends_with(&suffix)
+            });
+            let Some(target) = candidates.next() else {
+                continue; // no resolvable counterpart — stays its own row
+            };
+            if candidates.next().is_some() {
+                continue; // ambiguous — stays its own row
+            }
+            fold_target[index] = Some(target);
+        }
+
+        // Phase 3: emit surviving groups, then attribute folded orphan keys.
+        let mut report: Vec<ConflictReportRow<'_>> = Vec::new();
+        let mut report_index: Vec<Option<usize>> = vec![None; groups.len()];
+        for (index, group) in groups.iter().enumerate() {
+            if fold_target[index].is_some() {
+                continue;
+            }
+            report_index[index] = Some(report.len());
+            report.push(ConflictReportRow {
+                cache_key: group.representative,
+                state: group.state,
+                shadowed_keys: group.shadowed.clone(),
+            });
+        }
+        for (index, target) in fold_target.iter().enumerate() {
+            let Some(target) = *target else { continue };
+            let Some(position) = report_index[target] else {
+                continue;
+            };
+            report[position]
+                .shadowed_keys
+                .push(groups[index].representative);
+            report[position]
+                .shadowed_keys
+                .extend(groups[index].shadowed.iter().copied());
+        }
+        for row in &mut report {
+            row.shadowed_keys.sort_unstable();
+            row.shadowed_keys.dedup();
+        }
+        report
+    }
+
+    /// What an explicit duplicate-key migration would do to this cache
+    /// (TIN-3278). Pure: mutates nothing, writes nothing, takes no lock.
+    ///
+    /// Backs the dry-run half of the operator verb. Clones the entry map to
+    /// simulate the fold, so the plan and the apply step below can never
+    /// disagree — they run the identical routine.
+    pub fn plan_duplicate_key_migration(&self) -> KeyMigrationPlan {
+        let mut simulated = self.entries.clone();
+        fold_duplicate_keys(&mut simulated)
+    }
+
+    /// Fold duplicate key-namespace entries and persist the result (TIN-3278).
+    ///
+    /// **This is the only duplicate-key repair path.** It is deliberately not
+    /// reachable from `open`/`reload_from_disk`, because those are entered by
+    /// readers that hold no cross-process lock yet still write on drop when
+    /// `recovered_from_backup` is set — see the note in
+    /// [`StateCache::open`]. Repair is therefore an operation an operator
+    /// invokes, under the lock every other state-cache writer uses.
+    ///
+    /// `lock` is a witness, not decoration: it must be a [`StateFileLock`] for
+    /// *this* cache's path, and mismatches fail closed instead of being
+    /// trusted. The fold dirties the cache and flushes it while the caller's
+    /// guard is still alive.
+    ///
+    /// The fold's own invariant (see [`merge_duplicate_sync_states`]): the
+    /// surviving key's entry is authoritative for every value-bearing field, so
+    /// the only observable effects are that a duplicate key disappears, vector
+    /// clocks gain components, and `times_recorded` rises.
+    pub fn migrate_duplicate_keys_locked(
+        &mut self,
+        lock: &StateFileLock,
+    ) -> Result<KeyMigrationPlan> {
+        anyhow::ensure!(
+            lock.state_path() == self.db_path,
+            "refusing duplicate-key migration: held state lock is for {} but this cache is {}",
+            lock.state_path().display(),
+            self.db_path.display()
+        );
+
+        let applied = fold_duplicate_keys(&mut self.entries);
+        for record in &applied.merges {
+            tracing::warn!(
+                dropped_key = %record.dropped_key,
+                kept_key = %record.kept_key,
+                conflict_preserved = record.conflict_preserved,
+                "state cache: folded duplicate key-namespace entry (TIN-3278, explicit locked migration)"
+            );
+        }
+        for (key, reason) in &applied.skipped {
+            tracing::warn!(
+                key = %key,
+                reason = ?reason,
+                "state cache: duplicate key-namespace entry left in place (TIN-3278)"
+            );
+        }
+        if !applied.merges.is_empty() {
+            self.dirty = true;
+            self.flush().with_context(|| {
+                format!(
+                    "persisting duplicate-key migration: {}",
+                    self.db_path.display()
+                )
+            })?;
+        }
+        Ok(applied)
     }
 
     /// Capture a bounded set of cache entries so callers can roll back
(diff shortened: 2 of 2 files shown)


Jesssullivan:
## Final gate review — TIN-3278 — verdict: **request changes** (staying draft)

Reviewed at head `1a4dd83622284fc47a2cb8c981422c04499cbf65` (base `e7bfc1f0`). Tests were authored but never executed in the implement lane, so this round ran them on the designated build box (sting) and re-verified the review findings against the code.

**Net: the test suite is green and clippy is clean, but `cargo fmt --all --check` fails (a hard CI gate) and one substantive correctness gap remains. Not ready for review yet.**

---

### Sting run — `/tmp/tin3278-wt` @ `1a4dd836`, cargo 1.91.0

| Gate | Result |
| --- | --- |
| `cargo fmt --all --check` | **FAIL** (exit 1) — 2 diffs |
| `cargo test -p tcfs-sync` | **PASS** (exit 0) — 568 passed, 0 failed, 0 ignored |
| `cargo clippy -p tcfs-sync --all-targets -- -D warnings` | **PASS** (exit 0) — 0 warnings |

<details><summary><code>cargo fmt --all --check</code> — verbatim</summary>

```
=== CARGO_FMT_ALL_CHECK ===
Diff in /tmp/tin3278-wt/crates/tcfs-sync/src/state.rs:1837:
             .filter(|candidate| {
                 candidate.as_str() != key.as_str() && resolve_key_on_disk(candidate).is_some() && {
                     let normalized_candidate = crate::engine::normalize_rel_path_text(candidate);
-                    normalized_candidate.ends_with(&suffix) || normalized_candidate == normalized_orphan
+                    normalized_candidate.ends_with(&suffix)
+                        || normalized_candidate == normalized_orphan
                 }
             })
             .cloned()
Diff in /tmp/tin3278-wt/crates/tcfs-sync/src/state.rs:3899:
         };
         assert!(cache.mark_conflict(&file_path, conflict));

-        let known: std::collections::HashSet<String> =
-            ["neo".to_string(), "honey".to_string()].into_iter().collect();
+        let known: std::collections::HashSet<String> = ["neo".to_string(), "honey".to_string()]
+            .into_iter()
+            .collect();
         let ghosts = cache.conflicts_naming_unknown_devices(&known);
         assert_eq!(ghosts.len(), 1, "yoga is not in the known-device set");
         assert_eq!(ghosts[0].1.remote_device, "yoga");
FMT_EXIT=1
```
</details>

<details><summary><code>cargo test -p tcfs-sync</code> — the 6 new tests + per-target summaries</summary>

All six new tests pass:

```
test state::tests::tin3278_ambiguous_orphan_suffix_match_is_left_untouched ... ok
test state::tests::tin3278_merge_duplicate_sync_states_keeps_richer_conflict_and_max_times_recorded ... ok
test state::tests::tin3278_conflicts_naming_unknown_devices_reports_without_mutating ... ok
test state::tests::tin3278_migration_is_idempotent_on_second_load ... ok
test state::tests::tin3278_dedup_merge_never_drops_a_conflict_record ... ok
test state::tests::tin3278_dedup_merge_on_load_collapses_duplicate_key_namespaces ... ok
```

Per-target results (24 targets):

```
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 441 filtered out; finished in 0.00s
test result: ok. 442 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 13.09s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.51s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.70s
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.78s
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.22s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.75s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
TEST_EXIT=0
```

Totals: **568 passed; 0 failed; 0 ignored**. Zero warnings in the test build.
</details>

<details><summary><code>cargo clippy -p tcfs-sync --all-targets -- -D warnings</code> — verbatim tail</summary>

```
    Checking tcfs-chunks v0.12.17 (/tmp/tin3278-wt/crates/tcfs-chunks)
    Checking tcfs-storage v0.12.17 (/tmp/tin3278-wt/crates/tcfs-storage)
    Checking tcfs-sync v0.12.17 (/tmp/tin3278-wt/crates/tcfs-sync)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 50.40s
CLIPPY_EXIT=0
```

Zero clippy warnings emitted.
</details>

---

### Blocking items

#### 1. `open()` marking the cache dirty turns read-only CLI commands into unlocked writers

Confirmed by reading the call chain, not inferred:

- `open()` now sets `dirty: migration_dirty` (was `dirty: false`).
- `impl Drop for StateCache` (`crates/tcfs-sync/src/state.rs:1308-1316`) unconditionally flushes when `self.dirty || self.recovered_from_backup`.
- The `tcfs conflicts` offline branch (`crates/tcfs-cli/src/main.rs:9206-9210`) does `lock_explicit_state_cache(&state_path, state_override)` then `StateCache::open(...)`, and `state` is dropped at the end of that block.
- `lock_explicit_state_cache` (`crates/tcfs-cli/src/main.rs:1180-1189`) acquires `StateFileLock` **only when `state_override.is_some()`** — i.e. only for an explicit `--state <path>`.

So `tcfs conflicts` with no `--state` — the exact command in the ticket's 9-vs-8 repro — will now migrate, mark dirty, and rewrite `state.json` on drop **with no cross-process lock**, while the daemon flushes the same file under its own `StateFileLock`. Since `flush` finishes with an atomic rename, last-writer-wins can silently discard the daemon's concurrent conflict/state update. There are also several other unlocked `StateCache::open` sites in the CLI (`main.rs:4692, 4744, 4825`) that inherit the same new behavior.

Note `recovered_from_backup` already flushes unlocked today, so the pattern isn't unprecedented — but that is a rare recovery path, whereas this makes the *common* inspection path a writer.

Suggested direction: don't let load-time migration set `dirty`. Either keep the migration in-memory-only (fold for correct counting, persist nothing) and add an explicit lock-gated maintenance verb to durably rewrite, or have the migration persist only when the caller already holds a `StateFileLock`.

#### 2. `merge_duplicate_sync_states` discards one side's vector clock instead of joining it

This is the strongest correctness item, and it has an in-repo fix already sitting there. The merge picks the higher-`last_synced` side as `richer` and takes its `vclock` wholesale; `poorer.vclock` is dropped on the floor.

`VectorClock::merge` already exists and is exactly the right operation (`crates/tcfs-sync/src/conflict.rs:40-46`):

```rust
/// Merge another vector clock into this one (pointwise max).
pub fn merge(&mut self, other: &VectorClock) {
    for (device, &ts) in &other.clocks {
        let entry = self.clocks.entry(device.clone()).or_insert(0);
        *entry = (*entry).max(ts);
    }
}
```

Two entries being folded are by premise the same logical file, so the merged entry's clock should be the pointwise join of both — dropping one side erases causal history that `partial_cmp_vc` later relies on to classify divergence, which can turn a real concurrent pair into a false "dominates" (silently losing a conflict) on the next cycle. Please add `richer.vclock.merge(&poorer.vclock);` (and cover it with a test asserting the join, not the pick).

The same "richer wins wholesale" concern applies more weakly to `blake3`/`size`/`mtime`/`chunk_count`/`remote_path`/`device_id`: `last_synced` is a wall-clock field subject to skew, and an orphan's `last_synced` may simply be a frozen historical value that happens to be larger. Worth at least a comment justifying the tiebreak. The conflict `Option` handling itself is fine — see below.

---

### Non-blocking, but please address

**3. Pass 2's suffix heuristic mutates using a convention documented as read-only.** `get_by_rel_path`'s doc (`state.rs:975-981`) says cache keys are absolute paths that legitimately differ across hosts and that suffix matching exists to paper over that; pass 2 borrows the same convention but *merges*.

I want to be explicit that I checked this and it is **narrower than it first looks** — the broad "roamed cache gets erased" framing does not hold. For a genuine foreign-host key like `/home/otheruser/tcfs/notes.md`, pass 2 builds `suffix = "/home/otheruser/tcfs/notes.md"`, and a local `/home/jess/tcfs/notes.md` does **not** end with that, so no match and no merge. Likewise a locally-deleted file still resolves via its parent (pass 1, unchanged), and a deleted parent directory yields 0 candidates (left untouched). The reachable case is the intended one: a short legacy key like `/secrets/.audit.log` against exactly one `<root>/secrets/.audit.log`.

The residual real risk is multi-root: if a host has two registered roots sharing a relative suffix but the file is materialized under only one, pass 2 can fold a legacy key onto the wrong root's entry. Ambiguity guard (`candidates.len() != 1`) does not catch that because the other candidate isn't resolvable. A root-scoped match (only merge when the candidate lives under the same sync root the orphan was recorded for) would close it.

**4. Per-reload canonicalize cost.** `migrate_duplicate_keys` runs on every `reload_from_disk` (daemon calls it at `crates/tcfsd/src/grpc.rs:2326` and `4296`), doing an `fs::canonicalize` per tracked key each cycle — pass 2 is worse, calling `resolve_key_on_disk` again per candidate inside the filter, so an unresolved orphan costs O(keys) syscalls every cycle. Combined with item 1, every reload that still sees an unresolvable orphan re-marks dirty and re-emits the `tracing::warn!` line forever.

**5. Acceptance bar is met opportunistically.** Whenever an orphan has 0 or 2+ resolvable suffix candidates — the case the PR's own `tin3278_ambiguous_orphan_suffix_match_is_left_untouched` test constructs — migration correctly declines, and the `tcfs conflicts`-vs-daemon count divergence persists. This fixes the one duplicate observed live on neo, not the general class. Worth saying so in the PR body so the ticket isn't closed broader than the fix.

---

### Credit where due (verified sound)

- The conflict-specific merge logic is correct: a recorded `conflict` survives every merge path, `times_recorded` takes the max so a recurrence count can't regress, and `remote_manifest_key` is backfilled from the other side when missing — matching `mark_conflict`'s existing preserve-on-missing pattern. This held up under scrutiny.
- `resolve_key_on_disk` is a faithful fallback-free mirror of `path_key`, so pass 1 correctly leaves already-canonical and symlink-keyed entries alone rather than spuriously re-keying them.
- Pass ordering, the sorted-key snapshot for determinism, and the "restore what we took rather than lose it" guard in pass 2 are all good defensive choices.
- Root-cause analysis in the PR body matches the code: `conflicts()` filters raw entries with no dedup, so a duplicated pair double-counts.

---

### Interaction with #576 (TIN-3277) — reported only, no action taken

Checked as requested. **No textual conflict.** Trial merge of #576 head `6453eb7abfaa3294f97aa38f800194d99c2298de` into this branch's head, in a throwaway worktree:

```
$ git merge --no-commit --no-ff 6453eb7a
Automatic merge went well; stopped before committing as requested
$ git diff --name-only --diff-filter=U
(empty)
```

The merge was aborted immediately; nothing was pushed. #576 touches only `crates/tcfs-sync/src/reconcile.rs`, this PR only `crates/tcfs-sync/src/state.rs` — disjoint files, no construct-level overlap. Both are `MERGEABLE` against `main` at the same base. Landing order is unconstrained by text; whichever lands second just needs a normal CI pass.

---

### What the next round needs

1. Stop load-time migration from marking the cache dirty on unlocked paths (item 1) — this is the one that can lose operator data.
2. Join the vector clocks with the existing `VectorClock::merge` instead of picking one (item 2), plus a test on the join.
3. Run `cargo fmt --all` to clear the two mechanical diffs.
4. Optional but cheap: root-scope the pass-2 match (item 3), and narrow the PR body's claim to the single duplicate actually fixed (item 5).

Re-run on sting after those: `cargo fmt --all --check`, `cargo test -p tcfs-sync`, `cargo clippy -p tcfs-sync --all-targets -- -D warnings`. Test and clippy are already green, so items 1-3 are the whole remaining gap.

Jesssullivan:
## v2 — round-1 findings → fixes

Reworked at head `bbdf6694c5caed63ce003dc507bb4dcc0564ff3c` (was `1a4dd836`,
force-with-lease over it). Still **draft**, still a single-file diff
(`crates/tcfs-sync/src/state.rs` only).

| # | Round-1 finding | Disposition |
| --- | --- | --- |
| 1 | vclock merge drops causal history | **FIXED** — clocks joined with the existing `VectorClock::merge`; 2 new tests |
| 2 | unlocked-write hazard via `dirty` on migration | **FIXED (option b)** — migration is in-memory only, never dirties; `debug_assert` + byte-identical-file test |
| 3 | `cargo fmt --all --check` fails (2 diffs) | **FIXED** — fmt clean on sting |
| 4 | multi-root residual risk; keep sound parts unchanged | **DOCUMENTED** in the PR body and on `migrate_duplicate_keys`; sound parts untouched |
| 5 | coordinate with #576's `tracked_is_exact` guard | **NOTED** in the PR body; trial merge re-verified clean |

---

### 1. Vector clocks are joined, not picked — FIXED

`merge_duplicate_sync_states` now calls the existing pointwise-max
`VectorClock::merge` (`conflict.rs:40-46`) instead of taking one side's clock:

- `richer.vclock.merge(&poorer.vclock)` — the merged entry's clock;
- inside a merged conflict payload, `keep.local_vclock.merge(&other.local_vclock)`
  and `keep.remote_vclock.merge(&other.remote_vclock)` — **same-side only**.
  A local clock is never folded into a remote clock; that would erase the very
  divergence the record describes. (`conflict_git` already resolves with
  `local ∨ remote`, so a same-side join only ever widens the resolved clock.)

One correction to the round-1 framing, worth recording because it changed the
test I wrote. Dropping a component cannot make *our* side falsely dominate —
`partial_cmp_vc` reads a missing component as `0`, so a dropped clock can only
ever lower us. The real loss runs the other way: against a peer that has
advanced on the surviving device, a genuinely **Concurrent** pair reads as
"**the remote** dominates" → the engine classifies `RemoteNewer` and silently
overwrites the dropped side's divergence with no conflict recorded. Same
outcome the finding names (a real conflict silently lost), opposite direction.
My first draft of the test asserted the wrong direction and **failed on sting**
(`left: None, right: Some(Greater)`); the committed test asserts the direction
that is actually reachable.

`tin3278_merge_joins_vector_clocks_instead_of_dropping_one_side`:

- dropped side carries `yoga:7`, kept side carries `neo:4` → merged entry keeps
  **both**;
- against peer `{neo:5}`: joined clock → `None` (honest Concurrent);
- kept-side-only clock → `Some(Less)` — the false remote dominance, pinned;
- join is order-independent (`merge(a,b).vclock == merge(b,a).vclock`).

`tin3278_merge_joins_conflict_side_clocks_same_side_only` covers the payload
joins, including `assert_ne!(conflict.remote_vclock.get("neo"), 3)` — a
local-only observation must not surface as a remote one.

### 2. Unlocked-write hazard — FIXED, option (b), and why

**Migration is in-memory only and never marks the cache dirty.**

The round-1 chain is exactly right: `Drop` flushes on `dirty`
(`state.rs:1308-1316`), and `lock_explicit_state_cache`
(`tcfs-cli/src/main.rs:1180-1189`) takes `StateFileLock` **only** when
`state_override.is_some()`, so `tcfs conflicts` with no `--state` held no lock
while the old code dirtied on migration.

I picked (b) over (a) "lock the default path too" because the existing lock
discipline puts `StateFileLock` at the *call site that intends to write* — the
CLI's `lock_explicit_state_cache`, the daemon around its own flush — never
inside `StateCache::open`. Acquiring a cross-process lock implicitly in a
constructor would make every read-only inspection path contend with (or
fail-closed against) the daemon, and `open()` has several non-writing callers.
Keeping `open()` non-mutating adds **no new writer anywhere**, which is the
conservative reading of "the migration pass must never introduce an unlocked
write."

Durability is not lost: the fold persists on the next flush by a legitimate
lock-holding writer. The daemon dirties on any `set`/`mark_conflict` and
`flush` serializes the whole map, so the folded map replaces the un-migrated
file under the daemon's own lock — no new verb needed.

Enforcement:

- `dirty: false` at construction + `debug_assert` carrying the invariant;
- the invariant is written out in the comment block in `open()` (naming the
  `tcfs conflicts` / `lock_explicit_state_cache` path explicitly) and repeated
  on `reload_from_disk`;
- `tin3278_migration_does_not_dirty_or_rewrite_for_unlocked_readers` — hermetic
  and does not need a race: open a cache holding a duplicate pair, assert the
  fold happened in memory (`len()==1`, one migration record) and `!dirty`, drop
  it (where a dirty flush would fire), then assert the primary file is
  **byte-identical** and still contains `"/secrets/.audit.log"`;
- `tin3278_migration_is_idempotent_on_second_load` was reworked accordingly: it
  now asserts the fold does not dirty, then stands in for the daemon with a
  `set` before `flush`, and only then asserts the orphan is gone from disk.

Two follow-ons from round-1's items 4/5 came free with this: `reload_from_disk`
no longer re-emits the same `warn` for a record already seen in-process (so a
disk-reintroduced orphan cannot log every cycle), and `migrate_duplicate_keys`
now does exactly **one** `fs::canonicalize` attempt per key per call, computed
up front and shared by both passes — pass 2 no longer re-resolves every
candidate per orphan.

### 3. `cargo fmt --all --check` — FIXED

Both diffs (the pass-2 filter condition at old `:1837` and the `HashSet`
collect at old `:3899`) are formatted; fmt is clean on sting.

### 4. Multi-root residual risk — documented, sound parts untouched

Documented in the PR body and as a **Known boundary** doc section on
`migrate_duplicate_keys`: the `candidates.len() != 1` guard does not catch a
multi-root collision, because the sibling root's candidate is not resolvable
and the guard therefore sees a single candidate. Root-scoped matching is the
close, and state-cache keys carry no root attribution today. Also stated
plainly that an orphan with 0 or 2+ resolvable candidates is left alone, so the
count divergence persists for that entry — the ticket should not be read
broader than the class fixed.

The confirmed-sound parts are byte-for-byte unchanged: conflict-`Option` merge,
`times_recorded` max, `remote_manifest_key` backfill, `resolve_key_on_disk`,
pass ordering, sorted-key determinism, restore-what-we-took guard. The scalar
tiebreak now carries the comment round-1 asked for: it is deliberate rather
than provably correct (`last_synced` is wall-clock, skew-prone), and the
safety-critical information is the order-independent clock join.

### 5. #576 coordination — noted, trial merge re-verified

The `tracked_is_exact` guard #576 v2 landed in `reconcile.rs` warns on
TIN-3278-style dup lookups; **this migration is the durable fix that guard is
waiting on** — it removes the duplicate namespace the guard can only report.
Stated in the PR body.

Re-verified disjointness at the new head, in the throwaway worktree, aborted
immediately:

```
$ git merge --no-commit --no-ff 6453eb7a
Automatic merge went well; stopped before committing as requested
$ git diff --name-only --diff-filter=U
(empty)
```

#576 touches only `reconcile.rs`; v2 is still `state.rs` only.

---

### sting run — `/tmp/tin3278-wt2` @ `bbdf6694`, worktree removed after

| Gate | Result |
| --- | --- |
| `cargo fmt --all --check` | **PASS** (exit 0) |
| `cargo test -p tcfs-sync` | **PASS** (exit 0) — 571 passed, 0 failed, 0 ignored |
| `cargo clippy -p tcfs-sync --all-targets -- -D warnings` | **PASS** (exit 0) — 0 warnings |

<details><summary><code>cargo fmt --all --check</code> — verbatim</summary>

```
=== CARGO_FMT_ALL_CHECK ===
FMT_EXIT=0
```
</details>

<details><summary><code>cargo test -p tcfs-sync</code> — TIN-3278 tests + per-target results, verbatim</summary>

```
   Compiling tcfs-sync v0.12.17 (/tmp/tin3278-wt2/crates/tcfs-sync)
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 444 filtered out; finished in 0.00s
test state::tests::tin3278_ambiguous_orphan_suffix_match_is_left_untouched ... ok
test state::tests::tin3278_dedup_merge_never_drops_a_conflict_record ... ok
test state::tests::tin3278_dedup_merge_on_load_collapses_duplicate_key_namespaces ... ok
test state::tests::tin3278_merge_duplicate_sync_states_keeps_richer_conflict_and_max_times_recorded ... ok
test state::tests::tin3278_merge_joins_conflict_side_clocks_same_side_only ... ok
test state::tests::tin3278_merge_joins_vector_clocks_instead_of_dropping_one_side ... ok
test state::tests::tin3278_migration_does_not_dirty_or_rewrite_for_unlocked_readers ... ok
test state::tests::tin3278_conflicts_naming_unknown_devices_reports_without_mutating ... ok
test state::tests::tin3278_migration_is_idempotent_on_second_load ... ok
test result: ok. 445 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 13.09s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 7 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.05s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.06s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.51s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.73s
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.77s
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.25s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.74s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 0 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.00s
TEST_EXIT=0
```

Totals: **571 passed; 0 failed; 0 ignored** across 24 targets (568 before, +3
new tests). Zero warnings in the test build.
</details>

<details><summary><code>cargo clippy -p tcfs-sync --all-targets -- -D warnings</code> — verbatim tail</summary>

```
    Checking tcfs-chunks v0.12.17 (/tmp/tin3278-wt2/crates/tcfs-chunks)
    Checking tcfs-storage v0.12.17 (/tmp/tin3278-wt2/crates/tcfs-storage)
    Checking tcfs-sync v0.12.17 (/tmp/tin3278-wt2/crates/tcfs-sync)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 54.07s
CLIPPY_EXIT=0
```
</details>

<details><summary>First-attempt failure, for the record</summary>

```
---- state::tests::tin3278_merge_joins_vector_clocks_instead_of_dropping_one_side stdout ----
thread '...' panicked at crates/tcfs-sync/src/state.rs:4036:9:
assertion `left == right` failed: pinning the hazard: the kept-side-only clock claims false dominance
  left: None
 right: Some(Greater)
test result: FAILED. 444 passed; 1 failed; 0 ignored; 0 measured; 0 filtered out; finished in 13.10s
```

That was my hazard construction being inverted (see item 1); the production
join was already correct. Fixed, re-run green above.
</details>

Jesssullivan:
## Round-2 gate: **reject** — stays DRAFT, round 3 needed

Reviewed head `bbdf6694c5caed63ce003dc507bb4dcc0564ff3c` (single-file diff, `crates/tcfs-sync/src/state.rs`). Line numbers below are PR-head numbering for `state.rs`; `conflict_git.rs` / `grpc.rs` numbers are `origin/main` (untouched by this PR).

### Genuinely fixed in v2

- **Round-1 #1 (vector-clock drop) — fixed, and the corrected hazard direction is right.** `VectorClock::merge` is a true pointwise max and `partial_cmp_vc` reads an absent component as `0`, so a dropped component cannot fabricate *our* dominance — it makes the **remote** falsely dominate (false `RemoteNewer`, silent overwrite). `merge_duplicate_sync_states` now joins entry clocks (state.rs:1777) and joins conflict-payload clocks strictly same-side (state.rs:1791-1792), with tests pinning `Some(Less)` on the kept-side-only clock, honest `None` on the joined clock, and the negative (`remote_vclock.get("neo") != 3`).
- **Round-1 #3 (lock deadlock) — not introduced.** Option (b) adds zero `StateFileLock` acquisitions.
- `resolve_key_on_disk` (state.rs:1722) is `path_key` minus the identity fallback, so pass-1 re-keying lands exactly on the key live `get`/`set`/`remove` derive. No symlink-divergence hazard.
- Gates were reported green (fmt / `cargo test -p tcfs-sync` 571 pass / clippy `-D warnings`). Not independently re-run here — and note that **all four blockers below are semantic and pass fmt, clippy, and this PR's own suite unchanged.**

### Must-fix (blocking)

**1. Conflict resurrection: a stale conflict is grafted onto a live clean entry, with no recency guard.**
`merge_duplicate_sync_states` state.rs:1796 `(None, Some(other)) => Some(other)` plus state.rs:1799-1801 `richer.status = FileSyncStatus::Conflict` attach the orphan's conflict payload to an entry that is currently `Synced`/clean, *even when the kept side's `last_synced` is later than the orphan conflict's `detected_at`* — i.e. the file demonstrably synced cleanly after that conflict was detected.

Pre-PR the orphan conflict was inert: `get`/`set`/`mark_conflict` all re-derive the key through `path_key`, so only raw scans (`conflicts()`, state.rs:1035) ever saw it. Post-PR it lands on the key the reconciler, FileProvider badges, and `conflict_git` actually use, and it becomes durable as soon as anything flushes.

The PR's own test `tin3278_dedup_merge_never_drops_a_conflict_record` (state.rs:3764) *is* the failure scenario: canonical entry `last_synced=2000`, `conflict=None`, `status=Synced`; orphan `last_synced=1000` carrying `detected_at=500`, `times_recorded=1881`. After `open()` the real file's entry is `status=Conflict` with an 1881-count payload — `tcfs conflicts` still reports 1, not 0, and the file is now *genuinely* conflicted rather than cosmetically double-counted. The test asserts that as correct, so it locks the defect in.

Required: drop (or park separately) a merged-in conflict whose `detected_at` predates the kept side's `last_synced`, and re-point that test.

**2. Round-1 #2 is only half fixed — `recovered_from_backup` still makes unlocked readers write the folded map.**
`flush()` short-circuits on `if !self.dirty && !self.recovered_from_backup` (state.rs:1105) and `Drop` fires on `if self.dirty || self.recovered_from_backup` (state.rs:1332-1334) — both unchanged from main. The fold runs at state.rs:824-825, *before* the struct is built, so `dirty: false` (state.rs:838) does not make it write-free: whenever `open()` recovered from `.bak` (corrupt primary at state.rs:766-787, or missing primary with a backup at state.rs:788-801) `recovered_from_backup` is true and `Drop` flushes the **migrated** map with no `StateFileLock` held.

The new invariant comment cites `lock_explicit_state_cache` (tcfs-cli/src/main.rs:1180-1189), whose own doc comment says verbatim: "`StateCache::open` may recover a missing or corrupt primary from backup and flush that recovery on drop, even for a nominally read-only command." That function exists *because of* the path this fix ignores. Before this PR that recovery flush wrote back content equivalent to what was read; now it writes a content-mutating fold (orphan key deleted, entries merged on a wall-clock tiebreak) from an unlocked reader racing the daemon's atomic-rename flush.

The `debug_assert!` at state.rs:845 cannot catch this — and it is vacuous regardless: `dirty: false` is a literal three lines above, so it asserts a compile-time constant, and it checks the wrong field. The regression test `tin3278_migration_does_not_dirty_or_rewrite_for_unlocked_readers` (state.rs:4099) writes only `state.json` and never a `.bak`, so it never exercises the recovery path — false assurance.

Required: skip the fold when `recovered_from_backup` is set (or gate the fold behind an explicit lock-holding caller), and add a `.bak`-recovery variant of the byte-identity test.

**3. `reload_from_disk` now breaks its own documented contract.**
The function is documented "Existing in-memory entries are NOT overwritten (in-memory wins)" (state.rs:904-905) and implements that with `self.entries.entry(key).or_insert(state)` (state.rs:912). Because the fold is deliberately non-dirtying, the orphan key stays on disk indefinitely, so every reload re-inserts it and then `migrate_duplicate_keys` (state.rs:931) folds it into the live entry again. That fold picks the higher-`last_synced` side for `blake3`/`size`/`mtime`/`remote_path`/`device_id` (state.rs:1768-1772) — a disk value overwriting a live in-memory value, exactly what the contract forbids.

Failure scenario: daemon has just re-hashed a locally edited file (in-memory `blake3=NEW`, `last_synced=1000` from a clock-skewed or legacy-dated write); disk orphan has `last_synced=5000`, `blake3=OLD`. Reload (the daemon does this per resolve request, `tcfsd/src/grpc.rs:4294-4296`) → orphan re-added → fold selects the orphan as "richer" → the live entry now claims `blake3=OLD`, and the next reconcile compares the file against a stale digest (skips the upload, or mislabels direction). The conflict graft from #1 also re-fires on every reload until some unrelated write flushes.

Required: either exclude `reload_from_disk` from the fold, or make the fold strictly non-overwriting for keys already present in memory.

**4. The grafted payload can hard-fail repo-group git resolution.**
`collect_repo_conflicts` (conflict_git.rs:1263-1291) iterates `state.conflicts()`, gates on `entry.status != FileSyncStatus::Conflict`, then `bail!`s the **whole** repo resolution when `entry.remote_path` or `conflict.remote_manifest_key` falls outside the selected storage prefix. v2 backfills `remote_manifest_key` from the other side (state.rs:1787-1789) and forces `status = Conflict` (state.rs:1800), so a legacy-prefix orphan's manifest key now rides onto a live in-repo cache key.

Failure scenario: an orphan carrying a pre-migration prefix folds onto a key inside a real repo → `collect_repo_conflicts` bails with "has a remote_manifest_key outside the selected storage prefix" → `tcfs resolve` for that entire repo group aborts. A one-file cosmetic double-count becomes a repo-wide resolution outage. The same graft also flips daemon routing at `tcfsd/src/grpc.rs:4304` (`Path::new(cache_key).starts_with(&git_dir)`), which previously never matched the non-repo orphan key and now can — silently widening which resolutions route into the git keep-both/bundle machinery. Neither behavior change has a test.

### Non-blocking, but please address while you're in here

- `conflicts_naming_unknown_devices` (state.rs:1052) has **zero callers** repo-wide (verified against `origin/main` and this head; the PR touches only `state.rs`). It is presented as addressing "TIN-3278 defect 2 — ghost device references," but nothing is wired to it, so that defect is not fixed. Being `pub` is the only reason clippy's `dead_code` lint stays quiet, so the green `-D warnings` gate is not evidence of use.
- "Cost: exactly one `fs::canonicalize` attempt per key per call" (state.rs:1838-1840) understates it. Pass 2 is O(orphans × entries) and rebuilds a fresh NFC-normalized `String` per candidate via `normalize_rel_path_text` (engine.rs:5116-5122 allocates per component, every call) — quadratic with allocation in the inner loop, on the `open()`/`reload` path. Separately, `open()` now issues one `canonicalize` syscall per key for every CLI invocation and daemon start, which it did not before.
- The `normalized_candidate == normalized_orphan` branch (state.rs:1926) is dead: candidates are absolute keys starting with `/`, while `normalized_orphan` has had leading slashes stripped by `trim_start_matches('/')` (state.rs:1914).
- Pass 2's suffix match is case-sensitive, but neo/honey are case-insensitive APFS. An orphan differing only in case will not fold, so "clears the live `/secrets/.audit.log` symptom" is filesystem-case-dependent and still unproven — consistent with the PR's own unchecked manual-verification box.
- `tin3278_migration_is_idempotent_on_second_load` only passes because it injects an artificial `cache.set(...); cache.flush()` between the two opens. That `set` is precisely what the production reader path does *not* do, so it demonstrates idempotence for a hypothetical lock-holding writer, not for the shipped behavior — the same gap that leaves must-fix #3's re-fold loop untested.
- Multi-root boundary is documented rather than fixed (state.rs:1830-1836); fine as scoped, but it wants its own ticket since state-cache keys carry no root attribution.

### What round 3 needs

1. Recency guard in `merge_duplicate_sync_states`: never graft a conflict whose `detected_at` predates the kept side's `last_synced` (park it instead), and re-point `tin3278_dedup_merge_never_drops_a_conflict_record` to assert the guarded behavior.
2. Make the fold genuinely write-free for unlocked readers: skip it when `recovered_from_backup` is set, or gate it behind a lock-holding caller. Replace the vacuous `debug_assert` with one that can fail, and add a `.bak`-recovery byte-identity test (corrupt primary + valid backup + duplicate pair → orphan still on disk after drop).
3. Make the fold non-overwriting for keys already resident in memory (or exclude `reload_from_disk` entirely), with a test: in-memory `blake3=NEW`/`last_synced=1000` vs disk orphan `blake3=OLD`/`last_synced=5000` → live entry keeps `NEW`.
4. Either don't force `status = Conflict` / backfill `remote_manifest_key` across the fold, or add tests for the two downstream consumers — `collect_repo_conflicts` prefix `bail!` and the `grpc.rs` `starts_with(&git_dir)` routing — proving the fold cannot widen either.
5. Wire `conflicts_naming_unknown_devices` to a caller or drop it from this PR (and drop the "defect 2 fixed" framing until it is wired).
6. Re-run the sting gates at the new head and keep the manual check on a *copy* of neo's live `state.json` — under `LAB_DEPLOY_FREEZE`, read-only copy only, no live state mutation.

No merge, no ready-for-review; PR remains DRAFT pending round 3.

Jesssullivan:
## v3 — round-2 findings → **architecture change**, not a sixth guard

Reworked at head `fee22353df9bf116828c2ec371e380e7652978e9` (was `bbdf6694`,
force-with-lease). Still **DRAFT**. Diff is now two files:
`crates/tcfs-sync/src/state.rs` plus a `crates/tcfs-cli/src/main.rs` subcommand.

### The decision, first

Round 1 found 2 defects in the load-time fold. Round 2 found 4 more. All six
share one cause — **mutating cache content on a path unlocked readers enter** —
so this round removes the load-time fold rather than guarding it a sixth time.

**Reporting collapses. Only an explicit, locked operation repairs.**

- `StateCache::conflicts_report()` — read-side collapse. `&self`, borrowed
  `SyncState`s verbatim, no merge / graft / re-key / delete. Zero new write
  paths.
- `StateCache::migrate_duplicate_keys_locked(&StateFileLock)` — the only
  mutator. `StateFileLock` now carries the cache path it guards, so the API
  verifies the caller's guard is for *this* cache and fails closed otherwise.
  Driven by `tcfs state-migrate-keys` (dry run by default).
- `open` and `reload_from_disk` are unchanged apart from comments recording why
  they must not repair.

The user-facing defect is a reporting defect — `tcfs conflicts` printing 9 where
the daemon reports `conflicts=8`. Nothing about that requires writing. Once
reporting no longer depends on repair, repair no longer has to be implicit.

I did weigh Option A (patch the fold against your six-point checklist) on the
merits. It closes all four, but leaves four live guards on an implicit write
path, each of which has to stay correct as the fold and its callers evolve, and
each of which is somewhere a round 4 can find a hole. Option B leaves one write
path a caller must ask for by name and hold a lock to reach. B has a structural
argument; A has four behavioral ones. Full comparison table in the PR body.

The cost, stated plainly: repair is no longer automatic. An operator runs
`tcfs state-migrate-keys --execute`. That is the trade.

### Per-finding disposition

| # | Round-2 must-fix | Disposition |
| --- | --- | --- |
| 1 | Conflict resurrection: stale conflict grafted onto a live clean entry; the PR's own test locked the defect in | **ELIMINATED** |
| 2 | `recovered_from_backup` still flushes the migrated fold unlocked; `.bak` never exercised by the test | **ELIMINATED** |
| 3 | `reload_from_disk` breaks its "in-memory wins" contract | **ELIMINATED** |
| 4 | Grafted legacy-prefix `remote_manifest_key` bails `collect_repo_conflicts`; flips `grpc.rs` git routing | **ELIMINATED** |
| 5 | `conflicts_naming_unknown_devices` unwired | **DROPPED** from the PR |

**#1 — eliminated, and more strongly than the recency guard you asked for.**
You asked to drop a merged conflict whose `detected_at` predates the kept side's
`last_synced`. Instead the fold **refuses outright** whenever only the duplicate
carries a conflict (`WouldResurrectConflict`) — at *any* `detected_at` — and
also when the duplicate's conflict is newer than the survivor's
(`DuplicateConflictIsNewer`, the "premise doesn't hold, an operator should
look" case). Both keys stay: no record lost, no clean entry conflicted.
Refusals surface in `KeyMigrationPlan::skipped`, which the operator reads in the
dry run before `--execute`. The test that locked in the defect is gone;
`tin3278_explicit_migration_refuses_to_resurrect_a_stale_conflict` asserts the
refusal, that both keys survive, and that the clean entry stays `Synced`.

The general rule behind it — the **fold invariant**:

> The surviving key (the key `path_key()` derives, the one a live cycle has been
> maintaining) is authoritative for every value-bearing field. The fold may only
> (1) join vector clocks pointwise, (2) raise `conflict.times_recorded`,
> (3) make the duplicate key disappear.

v2's higher-`last_synced` tiebreak — wall-clock and skew-prone, as you noted —
is gone entirely.

**#2 — eliminated structurally.** No fold on `open`, so the
`recovered_from_backup` flush writes back exactly what it read. You were right
that `dirty:false` was not the relevant question and that the old
`debug_assert` was vacuous — it asserted a literal three lines above itself and
checked the wrong field. Deleted.
`tin3278_bak_recovery_open_writes_back_unfolded_content` builds the shape you
specified: corrupt primary + valid `.bak` + duplicate pair → asserts
`recovered_from_backup` is set, drops the cache so the unlocked recovery flush
actually fires, then re-loads and asserts the duplicate key survived, no entry
merged, and the `.bak` source is byte-identical.

**#3 — eliminated structurally.** No fold on reload; `or_insert` is again the
whole contract. `tin3278_reload_from_disk_keeps_in_memory_values` runs your
exact scenario — in-memory `blake3=NEW`/`last_synced=1000` vs disk
`blake3=OLD`/`last_synced=5000` — and asserts the live entry keeps `NEW`.

**#4 — eliminated by the invariant.** No `remote_manifest_key` backfill, no
`status` forcing, so neither `collect_repo_conflicts`' prefix `bail!` nor
`grpc.rs`' `starts_with(git_dir)` routing can observe a value at a key that key
did not already carry. `tin3278_explicit_migration_cannot_widen_git_consumer_fields`
constructs a `.git`-internal pair whose stale duplicate carries a *higher*
`last_synced` and legacy prefixes in **both** `remote_path` and
`remote_manifest_key`, plus a clean/conflicted refusal pair, and asserts: the
conflicted-key set after ⊆ before; survivor's `status` / `remote_path` /
`remote_manifest_key` / `rel_path` / `blake3` unchanged; `times_recorded` rose to
1881; no folded key carries a `legacy/` address; refused duplicate untouched.

One honest edge I'll name rather than let you find: a pass-1 **pure** re-key
(destination absent) does move an existing conflict onto a key the cache did not
hold, which `starts_with(git_dir)` could then match. Not widening — the
destination is by construction the key `path_key()` derives for the same file
(same `file_name`, canonicalized parent), so the entry lands where a live
`get`/`set` was always going to look; routing follows the live reconciler
instead of diverging from it, which is the bug being repaired. Recorded on
`fold_duplicate_keys`.

**#5 — dropped.** `conflicts_naming_unknown_devices` and its tests are removed.
The "defect 2 / ghost devices" framing is out of the PR body. You were right
that `pub` was the only reason `-D warnings` stayed quiet.

### Your non-blocking items

- **Pass-2 cost** — eliminated from the hot path rather than optimized:
  `open`/`reload_from_disk` now issue **zero** extra `canonicalize` syscalls,
  because the migration does not run there. `conflicts_report` canonicalizes
  once per *conflict* row (the small minority of a cache), not per entry.
- **Dead `normalized_candidate == normalized_orphan` branch** — removed.
- **Idempotence test propped up by an injected `set`+`flush`** — repointed.
  `tin3278_explicit_migration_is_idempotent` folds under lock, immediately
  re-runs the same locked migration and asserts the plan is empty, then reopens
  the persisted file and asserts `plan_duplicate_key_migration()` is empty. No
  injected write.
- **Case sensitivity on APFS** — documented as a *declining* case, not claimed
  fixed. Matching is byte-exact; an orphan differing only by case is left in
  place and keeps double-counting.
- **Multi-root boundary** — still a boundary (keys carry no root attribution),
  but no longer silent: 2+ resolvable candidates surface as
  `NoUnambiguousTarget` in the plan the operator reads before `--execute`.
  `tin3278_explicit_migration_reports_ambiguous_multi_root_orphan` covers it.
- **Item 6, manual check on neo** — still unchecked, deliberately. Under
  `LAB_DEPLOY_FREEZE` this lane is source-only; it is left as an explicit
  read-only-copy step in the test plan rather than silently ticked.

Also worth flagging, since `conflicts_report()` is new and I did not wire it
everywhere: `tcfs conflicts --root <id>` is served by the daemon's
`conflict_records` (`tcfsd/src/grpc.rs`), which still scans `conflicts()` raw.
That path is prefix-validated per root by `validate_conflict_cache_route`, and
the observed 9-vs-8 divergence is on the primary/legacy cache the offline branch
reads — so wiring the daemon is a separate change in a separate crate,
deliberately not bundled. Named here so it is a scoping decision, not an
oversight.

### Gates — sting, `/tmp/tin3278-wt4` @ `fee22353`, cargo 1.91.0

Fresh worktree from the sting clone; removed after (`git worktree list | grep -c
tin3278` → `0`).

| Gate | Result |
| --- | --- |
| `cargo fmt --all --check` | **PASS** (exit 0) |
| `cargo test -p tcfs-sync` | **PASS** (exit 0) — 579 passed, 0 failed, 0 ignored |
| `cargo clippy -p tcfs-sync --all-targets -- -D warnings` | **PASS** (exit 0) — 0 warnings |
| `cargo clippy -p tcfs-cli --all-targets` | **PASS** — 0 diagnostics in `crates/tcfs-cli` (caveat below) |

<details><summary><code>cargo fmt --all --check</code> — verbatim</summary>

```
=== CARGO_FMT_ALL_CHECK ===
FMT_EXIT=0
```
</details>

<details><summary><code>cargo test -p tcfs-sync</code> — 17 TIN-3278 tests + totals</summary>

```
test state::tests::tin3278_conflicts_report_collapses_resolvable_alias_key ... ok
test state::tests::tin3278_conflicts_report_collapses_duplicate_rows_without_mutating ... ok
test state::tests::tin3278_conflicts_report_never_hides_a_lone_orphan_conflict ... ok
test state::tests::tin3278_explicit_migration_dry_run_plan_does_not_mutate ... ok
test state::tests::tin3278_conflicts_report_leaves_ambiguous_orphan_visible ... ok
test state::tests::tin3278_fold_joins_conflict_side_clocks_same_side_only ... ok
test state::tests::tin3278_explicit_migration_refuses_when_duplicate_conflict_is_newer ... ok
test state::tests::tin3278_explicit_migration_refuses_to_resurrect_a_stale_conflict ... ok
test state::tests::tin3278_fold_joins_vector_clocks_instead_of_dropping_one_side ... ok
test state::tests::tin3278_explicit_migration_requires_a_lock_for_this_cache ... ok
test state::tests::tin3278_explicit_migration_reports_ambiguous_multi_root_orphan ... ok
test state::tests::tin3278_open_never_folds_or_rewrites_duplicate_keys ... ok
test state::tests::tin3278_bak_recovery_open_writes_back_unfolded_content ... ok
test state::tests::tin3278_reload_from_disk_keeps_in_memory_values ... ok
test state::tests::tin3278_explicit_migration_folds_and_persists_under_lock ... ok
test state::tests::tin3278_explicit_migration_cannot_widen_git_consumer_fields ... ok
test state::tests::tin3278_explicit_migration_is_idempotent ... ok
```

Selected per-target results (24 targets, all `ok`; 579 passed in total, 0 failed,
0 ignored):

```
test result: ok. 453 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 13.10s
test result: ok. 13 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.78s
test result: ok. 15 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 12.74s
test result: ok. 11 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.04s
test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.25s
test result: ok. 8 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 4 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
test result: ok. 3 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 0.03s
TEST_EXIT=0
```
</details>

<details><summary><code>cargo clippy -p tcfs-sync --all-targets -- -D warnings</code> — verbatim tail</summary>

```
    Checking tcfs-chunks v0.12.17 (/tmp/tin3278-wt4/crates/tcfs-chunks)
    Checking tcfs-storage v0.12.17 (/tmp/tin3278-wt4/crates/tcfs-storage)
    Checking tcfs-sync v0.12.17 (/tmp/tin3278-wt4/crates/tcfs-sync)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 47.44s
CLIPPY_EXIT=0
```
</details>

<details><summary><code>cargo clippy -p tcfs-cli</code> — and why it needed a local dep pin</summary>

Stated because it is a weaker gate than the other three. `-p tcfs-cli` **cannot
be resolved on sting at all**:

```
error: rustc 1.91.1 is not supported by the following package:
  async-notify@0.3.4 requires rustc 1.92.0
BASE_CLI_EXIT=101
```

Verified pre-existing — that is the output from a clean `origin/main` worktree
on the same box, not from this branch. To typecheck the CLI change anyway it was
built with a **worktree-local** `cargo update async-notify --precise 0.3.2`.
`Cargo.lock` is **not** modified in this PR, and the three primary gates above
all ran against the pristine lockfile *before* that pin was applied.

```
    Checking tcfs-cli v0.12.17 (/tmp/tin3278-wt4/crates/tcfs-cli)
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 31.31s
RAW_EXIT=0
=== tcfs-cli diagnostics ===
(none)
```

The only diagnostic in that whole build is a pre-existing
`clippy::only_used_in_recursion` warning in `crates/tcfs-secrets/src/kdbx.rs`,
untouched by this PR.
</details>

Staying DRAFT for the round-3 gate.

Jesssullivan:
2026-08-15 provenance/CI hold: the sole commit `fee22353df9bf116828c2ec371e380e7652978e9` is GitHub-unverified/unsigned and this lane previously used force-with-lease. Hosted greens are non-authoritative; the hosted FUSE red was a runner capability mismatch. Keep this PR draft/unmerged. After signed #572→#565→replacement #576, replay the stable explicit locked cache-key repair once on a fresh signed branch, supersede #577, and obtain full source plus natural `tinyland-*` proof. No live migration, rebase/force, or rerun.

Jesssullivan:
**TIN-3278 round-3 adversarial gate — REJECT (third recorded rejection), 2026-08-22**

Round 3 (head `fee22353`, the Option-B read-side collapse/report plus explicit locked migrate verb) was adversarially re-reviewed under the takeover lane. Verdict: reject, three must-fixes. Recorded here because no prior record of this round exists on the PR or the tracker.

- **MF1 — `conflicts_report()` suppresses a real distinct conflict (confirmed by test).** Its phase-2 ambiguity guard counts candidates only among conflict rows (`crates/tcfs-sync/src/state.rs:1109-1230`), while `fold_duplicate_keys` counts candidates over ALL entries. A clean sibling entry in a second root is invisible to the report and visible to the fold, contradicting the PR's own doc claim that the two agree.
- **MF2 — fold pass 1 re-keys a legacy root-relative key onto an unrelated SYSTEM path (confirmed by test; worst of the set).** `resolve_key_on_disk` canonicalizes the key's parent unconditionally with none of pass 2's guards. Any sync root with a top-level `lib/`, `bin/`, `var/`, or `tmp/` can plan a "successful fold" that, under `--execute`, deletes the real entry and plants a `Synced` entry with live `remote_path`/`blake3` outside every sync root. The dry run prints it as a benign canonicalization.
- **MF3 — the Option-B safety premise is false.** `StateFileLock` does not exclude the daemon: tcfsd holds only `daemon-instance.lock` (`crates/tcfsd/src/daemon.rs:319-320`) and calls `cache.flush()` on its long-lived `Arc<Mutex<StateCache>>` at **22 non-test call sites** with no `state.json` lock anywhere in the daemon. The repair therefore un-sticks deterministically on the daemon's next flush while the CLI reports it applied, and the accompanying `last_nats_seq` rewind exposes NATS replay on restart.

Scope amendment, recorded explicitly per the 2026-08-15 anti-silent-rescope clause: the replacement replay is NOT the byte-stable v3 patch. It is Option B plus three guards — (1) the report's ambiguity denominator becomes all entries, matching the fold; (2) fold targets must resolve inside a registered sync root or the fold refuses; (3) the migrate verb additionally acquires `daemon-instance.lock`, so it fails closed while tcfsd runs. The alternative shape (a daemon-routed repair RPC executing under the daemon's own cache mutex) remains open to review at replacement time.

Sequencing unchanged from the standing 2026-08-15 order: this PR stays draft; supersede-and-close occurs only when the signed replacement draft exists, after #572 and #565 land. No merge, rebase, force-push, rerun, or live migration.

Related to TIN-3278

PR #590, opened 2026-08-29
docs(ops): correct TIN-1547's premise -- FileProvider is deliberately read-only (TIN-2853)

Jesssullivan:
**Finding, not implementation.** This PR closes the design half of **TIN-1547** by reporting that the ticket's premise is stale in a specific, load-bearing way, and recommends a re-scope rather than execution against its current acceptance criteria.

## The finding

TIN-1547 assumes a FileProvider that can already mutate the remote — its acceptance criteria ask for rename to be "copy/upload-before-delete with rollback or explicit unsupported behavior," plus badge/progress/recovery assertions. That was true as of the ticket's own "Current proven path as of 2026-05-21" list.

It stopped being true on **2026-07-16**, commit `5bc2175` ("feat(sync): harden trusted stable-root pipeline (TIN-2853)"), which made every Apple-facing mutation entry point explicitly read-only pending "exact version-token conditional publication."

Verified by reading `origin/main` directly this pass, not inferred from docs:

- `crates/tcfs-file-provider/src/direct.rs:20` defines `FILE_PROVIDER_READ_ONLY_ERROR`; `tcfs_provider_upload` / `tcfs_provider_delete` — the exact FFI entry points a rename would call — both hit `reject_file_provider_mutation()` unconditionally.
- `crates/tcfs-file-provider/src/grpc_backend.rs` carries the same constant.
- `swift/fileprovider/Sources/Extension/FileProviderExtension.swift`'s `modifyItem` and `deleteItem` are unconditional `NSFileProviderError(.cannotSynchronize)` stubs — there is no rename code path in the shipped extension to hold a rollback story at all.
- `CHANGELOG.md` `[Unreleased]` states the same in product terms.

## What this means for the acceptance criteria

- **Rename** — already satisfied in the "explicit unsupported" branch, and more strongly than intended: a deliberate fail-closed posture, not a gap waiting for a handler. Implementing copy/upload-before-delete against the FFI as it exists today would either not compile or would reopen exactly the gap TIN-2853 closed.
- **Badge/progress/recovery** — explicitly scoped out; there is nothing to have progress about while the surface is read-only.
- **PZM soak** — deferred per this tranche's instruction.

## Also recorded

- Real prior art: a local branch `facet6/dotgit-conflict-corruption-harness` (head `f9fb683`, itself an `origin/main` ancestor) contains a pre-TIN-2853 rename implementation with its own unhandled upload-succeeds-delete-fails rollback gap — worth fixing when write support returns, and worth an operator check on whether that branch is still expected to be in flight.
- A recommended rename + rollback shape for whenever the version-token protocol lands.

## Ask

Confirm whether TIN-2853's read-only posture is an intentional still-standing decision. If it is, **TIN-1547 needs a PM re-write, not more engineering against its current acceptance criteria.**

## Lane ownership

This PR owns only `docs/ops/tcfs-fileprovider-hardening-status-2026-08-28.md` (new file). No crates touched, no code changes.

### DIFF
--- a/docs/ops/tcfs-fileprovider-hardening-status-2026-08-28.md
+++ b/docs/ops/tcfs-fileprovider-hardening-status-2026-08-28.md
@@ -1,0 +1,116 @@
+# TCFS FileProvider post-M10 hardening: premise re-check (TIN-1547)
+
+Date: 2026-08-28
+Status: finding + forward design, not implementation. Written to close the
+design half of TIN-1547 after re-verifying its premise against current
+`origin/main` and finding it stale in a specific, load-bearing way.
+
+## The premise TIN-1547 was written against is no longer current
+
+TIN-1547 (last substantive update 2026-05-26, per Linear) assumes a
+FileProvider that can already mutate the remote: its acceptance criteria ask
+for rename to be "copy/upload-before-delete with rollback or explicit
+unsupported behavior," for badge/progress/recovery assertions, and for one
+longer desktop soak. That assumption was true as of the ticket's "Current
+proven path as of 2026-05-21" list (exact hydrate, evict/rehydrate,
+FileProvider mutation remote pull, FileProvider rename remote pull).
+
+It stopped being true on **2026-07-16**, commit `5bc2175` ("feat(sync):
+harden trusted stable-root pipeline (TIN-2853)"), whose own commit message
+says it closes "FileProvider fail-closed gaps found during PR #551
+recovery." That commit made every Apple-facing mutation entry point
+explicitly read-only:
+
+```rust
+// crates/tcfs-file-provider/src/direct.rs:20
+const FILE_PROVIDER_READ_ONLY_ERROR: &str =
+    "TCFS FileProvider is read-only until exact version-token conditional publication is available";
+```
+
+`tcfs_provider_upload` and `tcfs_provider_delete` (the exact C FFI entry
+points a rename implementation would call) both call
+`self.reject_file_provider_mutation()` unconditionally on `origin/main`
+today -- confirmed by reading `crates/tcfs-file-provider/src/direct.rs` and
+`crates/tcfs-file-provider/src/grpc_backend.rs` directly (not inferred from
+docs). The `CHANGELOG.md` `[Unreleased]` section states this in product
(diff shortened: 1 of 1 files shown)


Jesssullivan:
**Note on the `Build PDFs` red X — it is a repo-wide infra flake, not this PR's content.**

Two distinct causes hit this PR, neither related to the change:

1. **Tectonic cannot fetch `.sty` files from its remote bundle.** Different file each attempt — `bitset.sty` on one run, `rerunfilecheck.sty` on another — each preceded by repeated `warning: failure requesting "<file>" from network` lines and Tectonic's own "this most probably is not Tectonic's fault -- please check your network connection." PR #582, which is not mine, fails the same job with the same exit code 201.

2. **`docs.yml` uses `concurrency: group: pages` with `cancel-in-progress: true`, and the group is not keyed by ref.** Any two PRs with docs runs in flight cancel each other, so a burst of PRs leaves `Build PDFs` / `Check Links` showing red on the losers even though nothing about them failed. That accounts for the `CANCELLED` states seen here earlier.

Worth an operator ruling on (2) specifically — `group: pages-${{ github.ref }}` for pull-request events would stop PRs from cancelling each other while keeping the real single-deploy serialization for `main`. I have not changed it: it is out of this ticket's scope and affects every PR in the repo, so it should be someone's deliberate call rather than a drive-by edit from a release lane.

Jesssullivan:
Follow-up evidence that `Build PDFs` is repo-wide and not this PR: `docs.yml` history shows the same job failing across unrelated branches and, decisively, **on a `push` to `main`** (run of 2026-08-23T03:50:02Z, `main -> failure`). Recent `pull_request` failures include `lane4/pr-c-release-flow`, `tin1556/d4-slug-module`, and `docs/tin1556-adr-rulings-20260729` — none of them mine.

Three re-runs on this branch failed on three network fetches (`bitset.sty`, then `rerunfilecheck.sty` twice), while the sibling doc PRs #589 and #591 passed the identical job with the same kind of markdown. That pattern is a flaky remote bundle, not a content defect.

Recommend treating this check as known-flaky for merge purposes and opening a separate ticket to pin Tectonic's bundle (or vendor the `.sty` set) so `docs.yml` stops failing on `main`.

PR #591, opened 2026-08-29
feat(cli): interactive storage wizard + best-effort status verify for tcfs init (TIN-1425)

Jesssullivan:
Tractable CLI slice of **TIN-1425** (BIP-39 wizard + first-run UX). The macOS host-app modal, Linux systemd hint, TUI variant, and postinstall-smoke adoption remain open — `docs/ops/tcfs-init-wizard-design-2026-05-18.md` now carries an honest "what landed vs what's still open" section rather than reading as fully delivered.

## The problem

Every installer drops binaries and then expects the user to hand-write `config.toml` plus S3 credentials. `tcfs init` silently carried over whatever `config` already had — on a fresh install, almost always nothing.

## What lands

- **`run_storage_wizard`** — on interactive runs (not `--non-interactive`, not `--skip-config`) prompts for backend (SeaweedFS / generic S3-compatible / keep existing), endpoint, bucket, region, and HTTPS enforcement. Existing values are shown as editable defaults, so re-running `tcfs init --force-config` against a configured setup doesn't force a retype. A wizard failure is a warning that falls back to the existing config, never a hard failure.
- **Credentials** — optionally writes the standard `~/.aws/credentials` `[default]` profile, which `tcfs_secrets::parse_aws_credentials_file` already falls back to. Reuses that existing path instead of inventing a new credential format or requiring a SOPS ceremony for a first run. File is `0600` on unix.
- **`verify_status_after_init`** — best-effort `tcfs status`-equivalent check at the end. Advisory only; returns `false` on any failure to reach tcfsd (the normal case right after a fresh install) and just leaves the existing guidance in place.
- Two pure unit tests for the yes/no parser (`parse_yes_no` is split out from the prompt precisely so it is testable without stdin).

Plain `stdin`/`rpassword` prompts, matching the existing mnemonic-confirmation prompt. No new dependency — deliberately not adding `dialoguer`/`inquire`, whose API this pass could not verify against.

## Fixed before shipping

The draft wrote `~/.aws/credentials` **unconditionally, destroying any other profiles**, with only a printed note. That is a data-loss bug in a first-run wizard aimed at users most likely to already have AWS profiles configured. Now an existing file is copied to a timestamped `credentials.tcfs-backup-<unix_s>` sibling first, the backup path is printed, and the copy failing aborts the write rather than proceeding. An ini-preserving merge would be the nicer fix but is not something to land un-compiled.

## Verification boundary — please read

This was authored under a **no-local-compiler constraint** and has **not been compiled or run locally**. What was verified instead, by reading `origin/main`:

- every referenced symbol exists — `connect_daemon_without_session`, `StatusRequest`, `storage_ok`, `Duration`/`SystemTime` imports, `config.daemon.socket`, and all four `StorageConfig` fields touched;
- `rpassword` is already a workspace dependency of `tcfs-cli`;
- `tcfs_secrets::parse_aws_credentials_file` exists and is the fallback this reuses.

**The first CI `cargo check` / `cargo clippy` / `cargo test` run on this branch is the real correctness gate, not this description.** Please do not merge on green-by-assumption without CI having actually run.

## Lane ownership

This PR owns `crates/tcfs-cli/src/main.rs` (the `cmd_init` region) and `docs/ops/tcfs-init-wizard-design-2026-05-18.md` — coordinate before stacking commits.

### DIFF
--- a/crates/tcfs-cli/src/main.rs
+++ b/crates/tcfs-cli/src/main.rs
@@ -5035,6 +5035,201 @@ struct InitOptions<'a> {
     password: Option<String>,
 }
 
+// ── `tcfs init` interactive wizard (TIN-1425) ───────────────────────────────
+//
+// Boundary: authored under a no-local-cargo-build constraint (NEO FENCE) --
+// not compiled, run, or otherwise verified locally. Needs a CI
+// `cargo check`/`cargo test` pass before it can be trusted. Deliberately
+// uses plain stdin/rpassword prompts (matching the existing mnemonic-
+// confirmation prompt just below) rather than adding a new prompt-library
+// dependency (dialoguer/inquire) whose API this pass can't verify against.
+
+#[derive(Debug, Clone, PartialEq, Eq)]
+struct WizardStorageChoice {
+    endpoint: String,
+    bucket: String,
+    region: String,
+    enforce_tls: bool,
+}
+
+/// Reads a line from stdin with a shown default; empty input keeps the
+/// default.
+fn prompt_with_default(prompt: &str, default: &str) -> Result<String> {
+    if default.is_empty() {
+        print!("{prompt}: ");
+    } else {
+        print!("{prompt} [{default}]: ");
+    }
+    std::io::stdout().flush().context("failed to flush prompt")?;
+    let mut line = String::new();
+    std::io::stdin()
+        .read_line(&mut line)
+        .context("failed to read prompt input")?;
+    let trimmed = line.trim();
+    if trimmed.is_empty() {
+        Ok(default.to_string())
+    } else {
+        Ok(trimmed.to_string())
+    }
+}
+
+fn prompt_yes_no(prompt: &str, default_yes: bool) -> Result<bool> {
+    let hint = if default_yes { "Y/n" } else { "y/N" };
+    let answer = prompt_with_default(&format!("{prompt} [{hint}]"), "")?;
+    Ok(parse_yes_no(&answer, default_yes))
+}
+
+/// Pure parsing half of `prompt_yes_no`, split out so it's testable without
+/// stdin. An empty or unrecognized answer keeps the caller's default rather
+/// than erroring -- a first-run wizard should never hard-fail on a typo.
+fn parse_yes_no(answer: &str, default_yes: bool) -> bool {
+    match answer.trim().to_lowercase().as_str() {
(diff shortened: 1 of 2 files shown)


Jesssullivan:
**Update — the compile gate has now run, and it caught something.**

`Build + Lint + Test` failed on the first run with five `cargo fmt --all -- --check` diffs (job `99032313532`). Those are now applied verbatim from the CI output (commit `c0f33b9`) and the job is **green**: it compiles, clippy passes, and the suite passes including the two new unit tests.

So the "not compiled locally" caveat in the PR body above is now stale in the reader's favour — but the *real* remaining gap is unchanged and worth stating plainly:

**Green does not cover the wizard.** Nothing in the suite exercises the interactive prompt sequence, the `~/.aws/credentials` write or its pre-write backup, or the post-init status check. `parse_yes_no` is the only new logic with test coverage. A manual or `expect`-driven fresh-install smoke is still the gate before this is trusted on a real first run — please don't read the green check as more than "it builds and the pure-logic tests pass."

`docs/ops/tcfs-init-wizard-design-2026-05-18.md` has been updated to say exactly that instead of the old "not compiled" line (commit `65401f9`).

Jesssullivan:
**Note on the `Build PDFs` red X — it is a repo-wide infra flake, not this PR's content.**

Two distinct causes hit this PR, neither related to the change:

1. **Tectonic cannot fetch `.sty` files from its remote bundle.** Different file each attempt — `bitset.sty` on one run, `rerunfilecheck.sty` on another — each preceded by repeated `warning: failure requesting "<file>" from network` lines and Tectonic's own "this most probably is not Tectonic's fault -- please check your network connection." PR #582, which is not mine, fails the same job with the same exit code 201.

2. **`docs.yml` uses `concurrency: group: pages` with `cancel-in-progress: true`, and the group is not keyed by ref.** Any two PRs with docs runs in flight cancel each other, so a burst of PRs leaves `Build PDFs` / `Check Links` showing red on the losers even though nothing about them failed. That accounts for the `CANCELLED` states seen here earlier.

Worth an operator ruling on (2) specifically — `group: pages-${{ github.ref }}` for pull-request events would stop PRs from cancelling each other while keeping the real single-deploy serialization for `main`. I have not changed it: it is out of this ticket's scope and affects every PR in the repo, so it should be someone's deliberate call rather than a drive-by edit from a release lane.

